PR #1, opened 2016-05-16
Implement config module

willkg:
This is similar in some ways to python-decouple, but it does things I needed it to do:

1. namespaces for configuration: makes it easier to have multiple components use configuration
2. configuration overrides for testing: makes it possible to test various configuration modes in tests
3. allows for adjusting configuration precedence: I'm not sure I need this for antenna, but it seemed generally useful

This supports .ini files and process environment configuration. It supports parsing strings into Python objects in a variety of different ways.

The one thing this doesn't do is automatic documentation generation. I'd really like to be able to point Sphinx at something that uses this configuration module and have it slurp all the configuration information and expose it in the docs without me having to do anything.

This is probably good enough for now.

r?

### DIFF
--- a/antenna/configlib.py
+++ b/antenna/configlib.py
@@ -1,0 +1,323 @@
+"""This module contains the configuration infrastructure allowing for deriving
+configuration from a .ini file and the process environment.
+
+In order of precedence:
+
+1. settings overrides (usually only applies when running tests)
+2. process environment
+3. .ini file (specified by ANTENNA_INI environment variable)
+
+Example of usage::
+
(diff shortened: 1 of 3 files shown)


willkg:
I have some questions about this:

1. Do we want to bother with `.ini` files or nix that code?
2. Do we want to support configuration in python modules like what Django does?
3. Do we think we're going to use this configuration system in other projects? In which case, I can extract it into a library.

willkg:
I'll re-create this with the right stuff.

PR #2, opened 2016-05-17
Implement config module

willkg:
This is similar in some ways to python-decouple, but it does things I
needed it to do:

1. namespaces for configuration: makes it easier to have multiple
   components use configuration
2. configuration overrides for testing: makes it possible to test
   various configuration modes in tests
3. allows for adjusting configuration precedence: I'm not sure I need
   this for antenna, but it seemed generally useful

This supports .ini files and process environment configuration. It
supports parsing strings into Python objects in a variety of different
ways.

r?


willkg:
There are a couple of questions:

1. Should we support `.ini` files?
2. Should we support Python module files similar to what we do with Django settings?
3. Do we think we're going to use this configuration system in other projects? In which case, I can extract it into a library. We can always do that later.

willkg:
@peterbe Can you review this? It's not used anywhere, but it should be reviewable.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+
+1. settings overrides (usually only applies when running tests)
+2. process environment
+3. .ini file (specified by ANTENNA_INI environment variable)
+
+Example of usage::
+
+    from antenna.configlib import config
+
+    DEBUG = config('DEBUG', default=True, parser='bool')
```
peterbe: That bothers me. It ought to be implicit from the default what the parser should be. It's like some old statically typed languages where you had to specify the type even though you give it a default. Like `var int n = 100`.

willkg: Two things:

First, this example is wrong. I need to fix it. It should be:

```
DEBUG = config('DEBUG', default='True', parser=bool)
```

The default is always the text version of the value--it's what you'd put in the environment, the ini file, etc.

Second, this structure makes it easier to encode complex values like db urls in the code. That way the parser has the job of breaking it out.

For example, what parser would this use?:

```
DATABASES = config('DATABASES', default='postgres://foo:bar@baz.com:5432')
```

peterbe: To answer your last question, that's a case where you supply your own custom parser. Or, you write:
```python
DATABASES = config('DATABASES', default='postgres://foo:bar@baz.com:5432')
import dj_database_url
DATABASES = dj_database_url.parse(DATABASES)
```

peterbe: I think `'True'` **looks** so like a mistake or something not very smart. 

The truth is, for booleans in a config system we probably want to do things like converting strings like `'no'`, `'0'` etc to `False`. 

What about simply...:
```
parser = type(default)
if parser is bool:
    value = str(value).lower() not in ('no', '0', 'false')
else:
    value = parser(value)
```
...or something like that.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+
+    @settings_override(FOO='bar', BAZ='bat')
+    def test_this():
+        ...
+
+
+Wait--what? Why'd write our own configuration library when there are so many
+out there?
+
+I wanted one that was flexible, supported namespaces and it needed to be easy
```
peterbe: Nit: Using first-person in checked in code makes it look less cooperative. "I" belongs in a personal blog post.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+
+    raise ValueError('%s is not a valid bool value' % val)
+
+
+def get_parser(parser):
+    """Returns a parsing function for a given parser"""
+    # Special case bool so that we can explicitly give bool values otherwise
+    # all values would be True since they're non-empty strings.
+    if parser is bool:
+        return parse_bool
```
peterbe: oh. my comment above was dumb. It helps to read before you speak.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+    true_vals = ('t', 'true', 'yes', 'y', '1')
+    false_vals = ('f', 'false', 'no', 'n', '0')
+
+    val = val.lower()
+    if val in true_vals:
+        return True
+    if val in false_vals:
+        return False
+
+    raise ValueError('%s is not a valid bool value' % val)
```
peterbe: I like this! That means that a typo like "not" doesn't accidentally get converted to `True` when I meant `False`.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+    """Override configuration layer for testing"""
+    def get(self, key, namespace=''):
+        if namespace:
+            key = '%s_%s' % (namespace, key)
+
+        key = key.upper()
+
+        for env in reversed(_CONFIG_OVERRIDE):
+            if key in env:
+                return env[key]
```
peterbe: neat!

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+        return NO_VALUE
+
+
+class ConfigDictEnv(object):
+    """dict-based configuration layer
+
+    Namespace is prefixed, so key "foo" in namespace "bar" is ``FOO_BAR``.
```
peterbe: Perhaps I don't understand the whole scope yet but what if I have something called `FOO_BAR` that is a key?

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+
+        # Add OS environment
+        envs.append(ConfigOSEnv())
+
+        # Add .ini environment
+        fn = os.environ.get('ANTENNA_INI', 'antenna.ini')
+        if not os.sep in fn:
+            fn = os.path.join(os.getcwd(), fn)
+        if os.path.exists(fn):
+            envs.append(ConfigIniEnv(fn))
```
peterbe: A missing `else` here might cause confusion. If I set `ANTENNA_INI=antenn.ini` (note typo) it's going to give up because it can't find `antenn.ini`.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+            use this differently
+        :arg default: the default value (if any)
+        :arg parser: the parser for converting this value to a Python object
+        :arg raise_error: True if you want a lack of value to raise a
+            ``ConfigurationError``
+
+        Examples::
+
+            # Use the special bool parser
+            DEBUG = config('DEBUG', default=True, parser=bool)
```
peterbe: Doesn't this mismatch with the general docstring now?

antenna/configlib.py
willkg: It is doing that. python-decouple does exactly the same thing.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+
+            from antenna.config_util import ListOf
+            ALLOWED_HOSTS = config('ALLOWED_HOSTS', default='localhost',
+                                   parser=ListOf(str))
```
peterbe: Neat!

antenna/configlib.py
willkg: I like explicit true and false values rather than "these are false and everything else is true". It keeps it more explicit.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+
+        # Return the default if there is one
+        if default is not NO_VALUE:
+            return parser(default)
+
+        # No value specified and no default, so raise an error to the user
+        if raise_error:
+            raise ConfigurationError(
+                '%s (%s) requires a value of type %s' % (key, namespace, parser)
+            )
```
peterbe: I don't understand the `else` situation here. What if I have a typo.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+        def test_foo():
+            ...
+
+
+    This can also be used as a context manager::
+
+        def test_foo():
+            with config_override(FOO='bar'):
+                ...
+
```
peterbe: Feel free to link to your blog post.

antenna/configlib.py
```diff
@@ -0,0 +1,323 @@
...
+    def __init__(self, **cfg):
+        self._cfg = cfg
+
+    def push_config(self):
+        _CONFIG_OVERRIDE.append(self._cfg)
+
+    def pop_config(self):
+        # Make sure it's not empty already because if it is, then we've done
+        # something horribly wrong.
+        assert _CONFIG_OVERRIDE
```
peterbe: You'd just get a nice `IndexError: pop from empty list` without this `assert`.

tests/data/config_test.ini
```diff
@@ -0,0 +1,5 @@
+[main]
+foo = bar
+
+[namespacebaz]
+foo = bat
```
peterbe: missing last newline.

tests/test_configlib.py
```diff
@@ -0,0 +1,115 @@
...
+    assert get_parser(str) == str
+    foo = lambda val: val
+    assert get_parser(foo) == foo
+
+
+def test_ListOf():
+    assert ListOf(str)('foo') == ['foo']
+    assert ListOf(bool)('t,f') == [True, False]
+    assert ListOf(int)('1,2,3') == [1, 2, 3]
+    assert ListOf(int, delimiter=':')('1:2') == [1, 2]
```
peterbe: Great test!

antenna/configlib.py
willkg: With the custom parser, you can do that right now:

```
DATABASES = config('DATABASES', default='postgres://foo:bar@baz.com:5432', parser=dj_database_url)
```

My point is that we can't guess what the parser should be based on the type of the default. In some cases it works, but not all. Given that, seems better to keep it consistent and make it always required.

peterbe:
>  Should we support .ini files?

Definitely! Needed for local development. 

>  Should we support Python module files similar to what we do with Django settings?

No need unless we need it :)

>  Do we think we're going to use this configuration system in other projects? In which case, I can extract it into a library. We can always do that later.

No, let's keep it pragmatically local so it can prove its worth.

peterbe:
This might be tangental but you what would be really cool, using `click` with the overriding ability to read from a `.ini` file and env. One of my favorite go projects is [flag](https://github.com/namsral/flag). It's like argparse/click but on steroids. It's a thin wrapper on Go's builtin argparse-equivalent with the ability to read from env first or `-config myfile.conf`.

peterbe:
Apart from nits (many of which are probably wrong or ignorable) r+. Go forth!

antenna/configlib.py
willkg: Mmm... That's true now. It used to be that I had the other things in the environment stack so it was possible for `.pop_config()` to pop off the process environment or the ini environment.

I'll nix this because you're right in that it's not needed plus it's terribly unhelpful.

antenna/configlib.py
willkg: Sure! It's possible that the confusion is coming from me using examples that aren't "rea". Let's use some real examples.

Say we have an antenna instance and we want it to add a crash id to the production rabbitmq component and also the stage rabbitmq component. The code is the same for the two rabbitmq components. They differ in configuration. They obviously have different queue names. So we could put the first one in the "prod" namespace and the second one in the "stage" namespace. Then we end up with configuration something like this:

```
from antenna.configlib import config

class RabbitMQComponent:
    def __init__(self, name):
        self.name = name
        self.queue = config('RABBITMQ_QUEUE', namespace=self.name, parser=str)
```

In the environment, you'd have these two keys:

```
PROD_RABBITMQ_QUEUE=prod_queue
STAGE_RABBITMQ_QUEUE=stage_queue
```

If you had an ini file for this configuration, you'd have this:

```
[main]
...

[prod]
RABBITMQ_QUEUE=prod_queue

[stage]
# This is for siphoning from prod to stage.
RABBITMQ_QUEUE=stage_queue
```

That's what I was thinking.

Does that make sense? Are there problems with it?

antenna/configlib.py
willkg: I mulled over what to do here. There are two conflicting use cases.

1. The code looks for an `antenna.ini` file, but I don't have one because I'm doing everything in the environment. In this case, I don't want it to error out.
2. The code looks for an ini file, but I've either put it in the wrong place or misspelled it in the ANTENNA_INI variable and so antenna doesn't pick it up and I'm getting weird configuration and antenna isn't telling me anything helpful to discover the problem.

I think both of these cases are likely to happen often. Thus I'm not sure what the best behavior here should be.

Further, it's likely that Python logging hasn't been set up, yet. So logging the error and moving on probably won't help.

What to do ...?

antenna/configlib.py
willkg: The module-level docstring is totally wrong. This is more correct.

antenna/configlib.py
willkg: I don't know what you mean. Can you show an example of the typo you're talking about?

willkg:
Thinking about command lines is interesting. This configuration system doesn't do anything with command lines. With this specific project, we're running things with gunicorn so I'm pretty sure we'd never use command line configuration. However if we use this for other projects, then we might want that. At a minimum, we might want an option to specify the ini file or at least have some infrastructure to make that easier. I think we can push that off until later, though.

willkg:
^^^ That fixes some of the review comments.

Outstanding issues:

1. What to do if the ini filename is typoed: https://github.com/mozilla/antenna/pull/2#discussion_r63569892
2. What you mean by typoing here: https://github.com/mozilla/antenna/pull/2#discussion_r63570493

I think that's it. Everything else seems good enough.

antenna/configlib.py
peterbe: Here's a radical idea; change it from `fn = os.environ.get('ANTENNA_INI', 'antenna.ini')` to `fn = os.environ.get('ANTENNA_INI', NO_VALUE)`.
It circumgoes the problem of whether or not to raise or swallow the error. 

Yeah, logging is not going to be useful. But `warnings.warn()` is an option. It goes to stdout by default which is almost always read when the presence of the `antenna.ini` file is expected to exist. I.e. in local development.

antenna/configlib.py
peterbe: Gee. I don't know what I mean. Can you explain to me, when do not you set `raise_error=True`?

If I understand this correctly, the only time you'd end up in this region of the code (where the `if raise_error` is) it's if no value was set in the env or in the `.ini` and there was no default. 
For example, in the of python-decouple where you often set `SECRET_KEY` without a default to force the user to set it. So when that's the case, why would you not want to raise errors?

antenna/configlib.py
peterbe: Excellent example! 
So it doesn't matter if the namespace has an `_` in it. Or the config key. There's little to gain from having to set the environment variable `PROD__RABBITMQ_QUEUE` instead.

antenna/configlib.py
```diff
@@ -0,0 +1,332 @@
...
+
+        # Go through environments in reverse order
+        for env in self.envs:
+            val = env.get(key, namespace)
+            if val is not NO_VALUE:
+                return parser(val)
+
+        # Return the default if there is one
+        if default is not NO_VALUE:
+            return parser(default)
```
peterbe: This is neat. At first I thought it would be weird to do `bool(True)` but things like that will always work. Like `str('defaultvalue')`

peterbe:
Not sure it matters greatly but if this config system ever makes it out of antenna to stand on its own two legs, the only thing that is odd is the hardcoded environment variable key `ANTENNA_INI`, right?
Is that something we should think about at this early stage?

antenna/configlib.py
```diff
@@ -0,0 +1,332 @@
...
+
+1. settings overrides (usually only applies when running tests)
+2. process environment
+3. .ini file (specified by ANTENNA_INI environment variable)
+
+Example of usage::
+
+    from antenna.configlib import config
+
+    DEBUG = config('DEBUG', default='True', parser=bool)
```
peterbe: I still don't understand what point of this is. Sorry. Being able to use `config('DEBUG', default=True)` feels a lot more pythonic. The parser can be confidently guessed by the type of the default value. 

And parser isn't just `bool` anyway. It needs to be that fancy function that can recognize 'yes', '1', etc. 

Also, it just feels like typing `'True'` is icky.

peterbe:
Loose question; should there be restrictions on the key? Like spaces or dots.

antenna/configlib.py
willkg: So, stepping back a bit and talking about "icky", "pythonic" and "feels"... these are words that imply subjective measures. That's cool. My experience has been that people use words like this when they can't argue in a more tangible way the pros and cons of a decision. In this case, I think this is Pythonic because it it creates an API that's explicit and simple in nature.

This API requires the programmer to explicitly specify the parser. That nixes a class of possible bugs and it's less code to implement and test. I don't think the tradeoff for adding code to infer the parser from the default here is good enough.

Regarding the default, I think making it always a string makes it consistent. It's *always* the string value that you would pass to the parser, no guessing required. Plus reading the code illuminates what kinds of values you can use in your ini file and os environment.

That seems like a good tradeoff to me for a configuration library that's trying to be the smallest one possible.

antenna/configlib.py
willkg: You could do other characters, too, as long as they're valid environment variables and ini file keys. I don't know offhand what the full set of those are. I was planning to just use alphanumeric characters and `_` for antenna.

antenna/configlib.py
peterbe: /me nods

antenna/configlib.py
willkg: There are two reasons to do this. First is that when calling `.config()` getting back a default value is different than getting back knowledge that the key wasn't set in config at all. With type systems, I really like having a separate "no value was provided" distinct from a default value. There are two cases I was thinking about where this helps.

Say you're writing a component and it uses some config value. Then many moons later, you go to change the component in a way that needs to be backwards compatible, but you want to support new behavior. So maybe you want to check for the new config value and if it's not there, then fall back to the old config value.

The second case is where a component has multiple modes each of which requires different configuration. You see this with components for Django's backends. Some backends require additional configuration that others do not. They try to gloss over this with things like `OPTIONS`, but it's still there. Being able to determine whether a key isn't set at all makes it easier to programmatically know which mode to do.

So, stepping back a bit... Does this configuration management system need this feature? I don't have a need for it in mind, but this kind of thing has bit me in the past and it seemed like a good thing to handle especially where the implementation is pretty straight-forward, easy to test and easy to document.

So, the second thing is whether to make `raise_errors=True` the default. I can't think of which default is better to have: raise errors or not. The `SECRET_KEY` use case is interesting and a compelling use case to make work by default. I'll change it.

antenna/configlib.py
willkg: I like this. If we require that `ANTENNA_INI` be set in the environment and points to a file, then we can throw a `ConfigurationError` if the file doesn't exist and not deal with any of the logging/warning stuff. I'll do that. 

The warnings module is also a good idea. I always forget about that.

willkg:
^^^ That has some more changes.

I really like this. The current result is a lot better than what I started with. Thank you!

antenna/configlib.py
```diff
@@ -20,6 +20,19 @@
...
+
+    SECRET_KEY = config('SECRET_KEY')
+
+
+If the ``SECRET_KEY`` is not provided, then this will raise a configuration
+error.
+
+If you specify a ``ANTENNA_INI`` environment variable with a path to your ini
+file, then that will be used. Otherwise, this only looks at the environment.
+
```
peterbe: Perhaps worth iterating that the environment always trumps the `.ini` file.

willkg: That's a good idea; I'll add it.

antenna/configlib.py
```diff
@@ -180,38 +193,49 @@ def _initialize_environments(self):
...
 
-        # Add .ini environment
-        fn = os.environ.get('ANTENNA_INI', 'antenna.ini')
-        if os.sep not in fn:
-            fn = os.path.join(os.getcwd(), fn)
+        # Add .ini environment if they specify ANTENNA_INI in the environment
+        fn = os.environ.get('ANTENNA_INI')
+        if fn:
+            if os.sep not in fn:
+                fn = os.path.join(os.getcwd(), fn)
```
peterbe: Not really a nit, but isn't that what `open()` does by default. If you type `open('file.ini')` it will implicitly convert that to `open(os.path.join(os.getcwd(), 'file.ini'))`. 
We could use this to our advantage and just try to read the file and if it doesn't exist, consider excepting an IOError. That could potentially kill two birds with one stone. I.e. you'd end up in the same IOError exception if the files doesn't exist OR can't be read.

antenna/configlib.py
```diff
@@ -180,38 +193,49 @@ def _initialize_environments(self):
...
+            else:
+                raise ConfigurationError(
+                    '%s does not exist.' % fn
+                )
 
         return envs
 
     def __call__(self, key, namespace=None, default=NO_VALUE, parser=str,
-                 raise_error=False):
+                 raise_error=True):
```
peterbe: A ha! This takes care of `SECRET_KEY = config('SECRET_KEY')` automatically.

tests/test_configlib.py
```diff
@@ -98,16 +99,31 @@ def test_ConfigIniEnv(datadir):
     assert cie.get('foo', namespace='namespacebaz') == 'bat'
 
 
+def test_config_ini_file_does_not_exist():
+    with mock.patch('os.environ') as os_environ_mock:
+        os_environ_mock.__contains__.return_value = True
+        os_environ_mock.get.return_value = 'doesnotexist.ini'
+        print os_environ_mock.mock_calls
+        print os_environ_mock.method_calls
```
peterbe: debugging left?

antenna/configlib.py
willkg: Regarding your first comment, you're right on. However, we change `fn` here so when we raise the configuration error later, it has the complete filename that was used. If we don't do that, then the IOError isn't helpful--the programmer is left wondering why the program can't find the ini file when it's *right there*!

Regarding your second comment, that's interesting.... How does it help to have both those cases be IOErrors rather than one a ConfigurationError and the other an IOError? Is there a situation where we'd want to catch these exceptions and do something more helpful?

tests/test_configlib.py
willkg: Rrrr... I removed it, but didn't save the file. Sorry about that.

As a side note, `mock_calls` is super helpful when you're trying to figure out wtf got called so you can mock it properly.

willkg:
^^^ Did a force push to nix the debugging.

I feel good about this. I'm going to land it in its current state so I can move forward. It's currently difficult to move forward without configuration.

Thank you so much for your diligence here. There are a couple of things that we differ on, but I totally know where you're coming from on them. I may change my mind as we start using it in "real life".

peterbe:
Regarding the idea of using an IOError exception to track file not existing on disk AND file not being readable. The exception is quite clear:
```
>>> open('foo.ini')
  File "<stdin>", line 1, in <module>
IOError: [Errno 2] No such file or directory: 'foo.ini'
```
And if you try to read a file you can't it's taken care of there too:
```
>>> import os
>>> os.path.isfile('/etc/sudoers')
True
>>> open('/etc/sudoers')
  File "<stdin>", line 1, in <module>
IOError: [Errno 13] Permission denied: '/etc/sudoers'
```

So arguably, you should not even have any `except ...:` clause. If the file has been *set* just read it and if it fails, let the `IOError` bubble up. 

This is highly academic and not important for the progress of the real work. But we're python geeks aren't we?

willkg:
Yeah... so that's exactly it. This is not clear to me which file it was looking at:

```
IOError: [Errno 2] No such file or directory: 'foo.ini'
```

What directory did it look in? How do I know what the current working directory is?

peterbe:
> What directory did it look in? How do I know what the current working directory is?

Yeah, I could totally see that being annoying.

PR #3, opened 2016-05-17
Add lint step to travis

willkg:
Reworks travis scaffolding so it runs with sudo and runs a lint pass as
well as a tests pass.


willkg:
Going to close and re-create this.

PR #5, opened 2016-05-17
Add lint step to travis

willkg:
Reworks travis scaffolding so it runs with sudo and runs a lint pass as
well as a tests pass.

Once more ... with meaning.


willkg:
The lint pass is working, but the tests are failing because it's missing a file I've got in another branch. I'm going to land this and move forward.

PR #24, opened 2016-05-20
Fix travis tests

willkg:
Travis uses pip 6 and so it doesn't jive with the hashes requirement.
This redoes things so that we move the test passes into tox as tox
environments rather than doing them in travis with the dispatch script.
This lets us more easily update pip and things like that and get a
better set of test runs.


willkg:
I should probably also add a "make the docs" pass, but I'm not going to right now. I need to fix that so it works right first.

willkg:
Hooray! Worked super! MERGING!

PR #28, opened 2016-05-25
Add ooid module and tests from socorrolib

willkg:
Antenna will be self-contained and won't depend on anything from socorro or socorrolib. However, there are a few things we need to use from socorrolib to make everything work.

As such, my plan is to copy code from socorrolib that's very very unlikely to change.

This copies over the ooid code and its tests. It also updates both so they don't make our linter sad and also to take advantage of pytest magic.

r?


willkg:
GitHub seems to be lagging and closed this out on me. So I created a new one.

PR #29, opened 2016-05-25
Add ooid module and tests from socorrolib

willkg:
Antenna will be self-contained and won't depend on anything from socorro or socorrolib. However, there are a few things we need to use from socorrolib to make everything work.

As such, my plan is to copy code from socorrolib that's very very unlikely to change.

This copies over the ooid code and its tests. It also updates both so they don't make our linter sad and also to take advantage of pytest magic.

r?


willkg:
Changes from the old code:

1. fix variable and function names
2. rewrite tests to use pytest assert stuff
3. minor code cleanup
4. better documentation

@peterbe Can you skim this for Python-correctness?

peterbe:
I really think you should take the opportunity to contemplate the naming. "ooid" is a relic of the past. "uuid" sounds like the standard python lib. "crashid" is tricky because I never know to write "crashid" or "crash ID" or "crashID".

antenna/lib/ooid.py
```diff
@@ -0,0 +1,112 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This is based on socorrolib's socorrolib/lib/ooid.py module.
+
+# OOID is "Our opaque ID"
+import datetime as dt
+import uuid as uu
+
+from antenna.lib.datetimeutil import utc_now, UTC
```
peterbe: Surely, the tests can't pass yet with this in them.

peterbe:
We need to solve the `datetimeutil` issue. From skimming the code it seems it needs `datetimeutil` for the sake of getting todays date (in UTC). Surely we can do with just standard python `datetime`. But that `UTC`? Is that something we can get from `datetime` too or do we need something that actually understands timezones like `python-dateutil` or `arrow`.

antenna/lib/ooid.py
```diff
@@ -0,0 +1,112 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This is based on socorrolib's socorrolib/lib/ooid.py module.
+
+# OOID is "Our opaque ID"
+import datetime as dt
```
peterbe: See chat in #broken.

tests/test_ooid.py
```diff
@@ -0,0 +1,118 @@
...
+import datetime as dt
+from unittest import TestCase
+
+from antenna.lib.datetimeutil import utc_now, UTC
+import antenna.lib.ooid as oo
+
+
+class Test_ooid(TestCase):
+    def setUp(self):
+        self.baseDate = dt.datetime(2008, 12, 25, tzinfo=UTC)
```
peterbe: `s/baseDate/base_date`  maybe?

antenna/lib/ooid.py
willkg: It's better for project archaeology and history to keep conversations in one place or another rather than have conversations diverge.

willkg:
So... outstanding issues here:

1. Requires datetimeutil. I had that queued for another PR, but I'll just toss it in here. I had forgotten ooid required that. I should have done them the other way around.
2. Change the ooid name. An ooid is not the same thing as a uuid, though in the case of the collector the ooid is only used for crash_ids. The code uses these two terms consistently, so I'm not sure there's an issue here.
3. Change import renaming in test file. I can do that.

willkg:
^^^ That cleans up the imports and `baseDate`. It also adds the datetimeutil module. That's a lot smaller than the one in socorrolib, because we're only using a very small part of it.

peterbe:
Regarding the name, I didn't have an opinion. I just wanted to poke you to stop and think about it. Sounds like you've done that. 
r+
Want me to merge it?

willkg:
Thank you!

PR #30, opened 2016-05-25
Modify config

willkg:
This tweaks configuration a little bit to make it easier to work with in the breakpad code I'm writing.

The only new feature is the `parse_class` parser. That should be pretty straight-forward.

The rest of it is nixing the config singleton for one that's instantiated when the app is created.

### DIFF
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -9,16 +9,26 @@ from functools import wraps
 
 import falcon
 
-from antenna.configlib import config
+from antenna.configlib import ConfigManager
 
 logger = logging.getLogger('gunicorn.error')
 
 
-def require_basic_auth(username, password):
+def require_basic_auth(fun):
     """Decorator for requiring HTTP basic auth
 
-    :arg username: the valid username to use
-    :arg password: the valid password to use
+    This is used in resources and requires the resource to have
+    ``is_valid_auth`` implemented.
+
+    Example::
+
+        class HealthCheckResource:
+            def is_valid_auth(self, username, password):
+                return (username, password) == ('foo', 'bar')
+
+            @require_basic_auth
+            def on_get(self, req, resp):
+                ...
 
     """
     def auth_error():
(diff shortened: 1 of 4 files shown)


willkg:
@peterbe Can you do a quick r? for python-y issues and a sanity check?

antenna/configlib.py
```diff
@@ -91,6 +101,22 @@ def parse_bool(val):
...
+
+    >>> parse_class('hashlib.md5')
+
+    """
+    module, class_name = val.rsplit('.', 1)
+    module = importlib.import_module(module)
+    try:
+        return getattr(module, class_name)
+    except AttributeError:
+        raise ValueError('%s is not a valid member of %s' % (
```
peterbe: Why a `ValueError`? What was wrong with the `AttributeError`? Also, might it be of interest to re-wrap it as `ImportError`?

antenna/app.py
```diff
@@ -9,16 +9,26 @@
 
 import falcon
 
-from antenna.configlib import config
+from antenna.configlib import ConfigManager
```
peterbe: I like it. It's simpler.

antenna/configlib.py
willkg: We could return the AttributeError, my initial tinkering suggested it wasn't wildly helpful and that a better error messages was. Without this wrapping, you get something like:

```
>>> import importlib
>>> module = importlib.import_module('hashlib')
>>> getattr(hashlib, 'foo')
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'foo'
>>> 
```

"module" is totally not helpful here. With my changes, you get something like:

```
ValueError: 'foo' is not a valid member of hashlib'
```

I thought about wrapping the ImportError and then decided not to because the ImportError itself is more informative than what I'd probably do if wrapping it.

willkg: Continuing, I think the actual exception is moot. This is something that the developer is seeing when they're implementing something and a user is seeing if they've got antenna configured wrong. It's not an exception that is intended to be caught and handled.

I'm on the fence. I think I want to hedge towards users being able to see the message and intuit what to do next. I don't know what to do to make it better.

With Python 3, this is much easier since you can chain exceptions. Then we could throw an additional `ConfigurationError` with a super helpful message and the user gets our message plus the original one.

peterbe:
r+

willkg:
Thank you!

PR #35, opened 2016-09-08
Dockerfile updated to be based on python, circle building properly

milescrabill:
update dockerfile to be based on python:2.7.12-slim, add circle.yml and boilerplate for building containers in circle

@willkg r+?

You can pull from dockerhub with `docker pull mozilla/antenna:unstable`. Once this is merged you can just `docker pull mozilla/antenna` and it will pull latest, which is built off of master.

:unstable will be built off of the latest non-master, non-tagged build (on any branch of _this_ repo)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,31 @@
-# Dockerfile for socorro-collector container
-FROM centos:centos7
+FROM python:2.7.12-slim
+
+WORKDIR /app/
+RUN groupadd --gid 1001 app && useradd -g app --uid 1001 --shell /usr/sbin/nologin app
+
+RUN apt-get update && apt-get install -y \
+    gcc apt-transport-https
+
+COPY ./requirements.txt /app/requirements.txt
+COPY requirements-dev.txt /app/requirements-dev.txt
+
+RUN pip install -U 'pip>=8' && \
+    pip install --upgrade --no-cache-dir -r requirements.txt
+
+# Install the app
+COPY . /app/
 
 # Set Python-related environment variables to reduce annoying-ness
 ENV PYTHONUNBUFFERED 1
 ENV PYTHONDONTWRITEBYTECODE 1
+ENV PORT 8000
 
-EXPOSE 8000
+USER app
+EXPOSE $PORT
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,28 +1,31 @@
...
 
-# Install pip and get pip8
-RUN yum -y install epel-release \
-    && yum -y update \
-    && yum -y group install "Development Tools" \
-    && yum -y install python-devel python-pip
-COPY bin/pipstrap.py bin/pipstrap.py
-RUN ./bin/pipstrap.py
+RUN pip install -U 'pip>=8' && \
+    pip install --upgrade --no-cache-dir -r requirements.txt
```
willkg: Looks like we're copying both the requirements files, but we're only installing the production one. That's fine for non-dev environments, but for development, we need the requirements-dev.txt one, too.

How should that work? Can we install dev requirements "only if" the container is for dev purposes?

Dockerfile
```diff
@@ -1,28 +1,31 @@
...
+USER app
+EXPOSE $PORT
 
-RUN pip install --require-hashes -r requirements-dev.txt \
-    && pip install -e .
+CMD ANTENNA_INI=settings_dev.ini gunicorn \
+    --workers=1 \
+    --worker-connections=4 \
+    --worker-class=gevent \
+    --bind localhost:${PORT} \
```
willkg: I usually bind to 0.0.0.0 which means to bind to all the addresses--including the loopback one. By default, gunicorn binds just to 127.0.0.1 which is the problem I usually run into.

You mentioned that you do some docker thing such that it works. Is that done somewhere in this PR that I'm not aware of? Or is it something I set up in docker on my local machine?

Dockerfile
willkg: I talked with @mythmon and I understand what's going on and I'm cool with this now.

Dockerfile
willkg: As long as you think this works in prod, I'm cool with this, too.

willkg:
This is the best! r+

PR #36, opened 2016-09-08
Fix docker infrastructure

willkg:
This fixes the Dockerfile to bind to 0.0.0.0:8000 because that makes
things work. This updates the Makefile and docker-compose.yml so we
can launch Antenna for development purposes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,12 +41,6 @@ htmlcov/
 coverage.xml
 *,cover
 .hypothesis/
-
-# Translations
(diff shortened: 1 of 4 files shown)


willkg:
@miles Does the Dockerfile changes look ok?

milescrabill:
lgtm!

willkg:
Circle is sad, but for other reasons. The tests pass. So I'm landing this now.

PR #38, opened 2016-09-09
Fix py.test sys.path issues

willkg:
This fixes the py.test path issues for good! Down with py.test issues! Up with CI!


willkg:
Yay! De-toxed and it works. Merging this.

PR #46, opened 2016-09-19
Don't deploy for dev branches, PRs, etc.

milescrabill:
Test commit, don't merge yet ;)


willkg:
Circle seems to have failed, but github is showing it as still running. The failure is entirely my fault and fixed in another PR.

Given that, I'm mergifying this now!

PR #47, opened 2016-09-22
WIP: Initial crash storage code, NoOpCrashStorage, and logging

willkg:
This adds some rough crashstorage scaffolding and a NoOpCrashStorage
that logs a line about the crash--it doesn't store anything anywhere.

This also fixes logging so that it works with gunicorn.


willkg:
I expect logging will change when we start putting this into a -stage and -prod environments, so I'm treating these changes as a stop gap.

The `NoOpCrashStorage` makes it easier for testing and development since it doesn't require anything and doesn't leave anything behind.

I'm going to do some other stuff, then come back and review this.

antenna/app.py
```diff
@@ -19,7 +20,54 @@
...
+            'level': 'WARNING',
+        },
+        'loggers': {
+            'antenna': {
+                'propagate': False,
+                'handlers': ['console'],
+                'level': 'DEBUG',
+            },
+        },
+    }
```
willkg: I lifted this from Normandy and did some minor changes. When Antenna becomes ops-ified, we'll likely change how logging is configured and how it works. Given that, I figured I'd do a "good enough first pass that I can get stuff done" pass and defer any more thought until later.

This has logging messages getting to gunicorn and being spit out to the console.

antenna/app.py
```diff
@@ -37,7 +85,7 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
 
     required_config.add_option(
         'crashstorage_class',
-        default='antenna.external.boto.crashstorage.BotoS3CrashStorage',
+        default='antenna.external.crashstorage_base.NoOpCrashStorage',
```
willkg: Figured I'd make the default be `NoOpCrashStroage`. In Socorro-land, we like to make the default whatever production uses. I see the usefulness of that, but in Socorro-land, we have no ability to run a prod-like setup whereas with Antenna, we'll have a dockerized prod-like setup. So I claim I'm solving the underlying need, but in a different way.

We can redecide that later if it proves to be an issue.

antenna/external/boto/crashstorage.py
```diff
@@ -8,4 +8,5 @@ def __init__(self, config):
         pass
 
     def save_raw_crash(self, raw_crash, dumps, crash_id):
-        print((raw_crash, dumps.keys(), crash_id))
+        # FIXME(willkg): Implement this
+        pass
```
willkg: I'll implement this in a later PR.

willkg:
TODO: Write a tiny integration test using the `NoOpCrashStore` just to prove that a HTTP POST makes it all the way to crash storage with the right shape.

willkg:
^^^ That rebases against master picking up the `/__version__` endpoint implementation. It also adds an integration test for crash storage.

willkg:
Circle is happy. I think this looks good. I'm going to land it.

PR #50, opened 2016-09-22
Implement the /__version__ endpoint

willkg:
This implements one of the endpoints that Ops wants in order to run
Antenna in production.

This also removes the previous iteration of the HealthCheckResource.

### DIFF
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -5,8 +5,8 @@
 import cgi
 import hashlib
 import io
-import json
 import logging
+import os
 import time
 import zlib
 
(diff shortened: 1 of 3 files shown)


antenna/app.py
```diff
@@ -193,30 +193,44 @@ def on_post(self, req, resp):
...
 
 def get_app(config=None):
     """Returns AntennaAPI instance"""
     if config is None:
         config = ConfigManager([ConfigOSEnv()])
+
+    basedir = config(
+        'basedir',
+        default=os.path.abspath(os.path.dirname(__file__))
+    )
```
willkg: I should hoist this up into an AppConfig so that we can more-better document app-level configuration. I'll write up an issue for that now, but do it later.

willkg:
The one thing I don't know for sure is if the default place this looks for `version.json` works with how we have things built on -dev. I'll find out after landing this and talk with @milescrabill about it.

PR #54, opened 2016-09-26
Implement FSCrashStore

willkg:
This implements a rudimentary file system crash store that mimics the
"directory structure" of what we're going to do on S3, but does it on
a local file system.

There are a bunch of FIXMEs that should get thought through at some
point, but since I'm only using this crash store for
development/testing purposes, I'm going to leave those be for now.

### DIFF
--- a/antenna/external/fs/crashstorage.py
+++ b/antenna/external/fs/crashstorage.py
@@ -55,9 +55,10 @@ class FSCrashStorage(RequiredConfigMixin):
     def __init__(self, config):
         self.config = config.with_options(self)
 
-        self.root = os.path.abspath(self.config('fs_root'))
-        # FIXME(willkg): We should probably do more to validate fs_root. Remove
-        # / at the end? Verify we can write files to it?
+        self.root = os.path.abspath(self.config('fs_root')).rstrip(os.sep)
+
+        # FIXME(willkg): We should probably do more to validate fs_root. Can we
+        # write files to it?
         if not os.path.isdir(self.root):
             os.makedirs(self.root)
 
@@ -76,6 +77,7 @@ class FSCrashStorage(RequiredConfigMixin):
         )
 
     def _get_dump_names_path(self, crash_id):
+        """Returns path for where the dump_names list should go"""
         return os.path.join(
             self.root,
             self._get_crash_date(crash_id),
@@ -84,6 +86,7 @@ class FSCrashStorage(RequiredConfigMixin):
         )
 
     def _get_dump_name_path(self, crash_id, dump_name):
+        """Returns path for a given dump"""
         return os.path.join(
             self.root,
             self._get_crash_date(crash_id),
@@ -101,8 +104,6 @@ class FSCrashStorage(RequiredConfigMixin):
         :arg crash_id: The crash id as a string.
 
         """
-        print((raw_crash, dumps.keys(), crash_id))
-
         files = {}
 
         # Add raw_crash to the list of files to save.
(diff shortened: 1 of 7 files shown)


willkg:
This crash store is used solely for development and testing at the moment. Given that, I'm going to walk away for a bit, let Circle do it's thing and then come back and self-review.

antenna/external/fs/crashstorage.py
```diff
@@ -0,0 +1,132 @@
...
+
+    """
+    required_config = ConfigOptions()
+    required_config.add_option(
+        'fs_root',
+        default='/tmp/antenna_crashes',
+        doc='path to where files should be stored'
+    )
+
+    # FIXME(willkg): umask
```
willkg: The `FSTemporaryStorage` has a umask configuration option, but I didn't add one here. Having one is probably a good idea if we ever "went production" with this.

antenna/external/fs/crashstorage.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __init__(self, config):
+        self.config = config.with_options(self)
+
+        self.root = os.path.abspath(self.config('fs_root'))
+        # FIXME(willkg): We should probably do more to validate fs_root. Remove
+        # / at the end? Verify we can write files to it?
```
willkg: TODO: Add some code to remove any trailing `/`.

antenna/external/fs/crashstorage.py
```diff
@@ -0,0 +1,132 @@
...
+        if not os.path.isdir(self.root):
+            os.makedirs(self.root)
+
+    def _get_crash_date(self, crash_id):
+        # NOTE(willkg): This assumes we're in the 21st century and that the last
+        # six characters of the crash_id are YYMMDD.
+        return '20' + crash_id[-6:]
```
willkg: The `antenna.lib.ooid` module has some functions for manipulating crash ids and this should probably be there, but those functions return `datetime`s and that's not what I want. So I could adjust those or add another function, but I figured I'd see if I needed to use that anywhere else first and if so, then centralize it.

antenna/external/fs/crashstorage.py
```diff
@@ -0,0 +1,132 @@
...
+        """Returns path for where the raw crash should go"""
+        return os.path.join(
+            self.root,
+            self._get_crash_date(crash_id),
+            'raw_crash',
+            crash_id + '.json'
+        )
+
+    def _get_dump_names_path(self, crash_id):
+        return os.path.join(
```
willkg: TODO: Add docstring.

antenna/external/fs/crashstorage.py
```diff
@@ -0,0 +1,132 @@
...
+            self.root,
+            self._get_crash_date(crash_id),
+            'dump_names',
+            crash_id + '.json'
+        )
+
+    def _get_dump_name_path(self, crash_id, dump_name):
+        return os.path.join(
```
willkg: TODO: Add docstring.

antenna/external/fs/crashstorage.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # Save all the files.
+        for fn, contents in files.items():
+            logger.debug('Saving file %r', fn)
+            path = os.path.dirname(fn)
+
+            # FIXME(willkg): What happens if there is something here already and
+            # it's not a directory?
+            if not os.path.exists(path):
+                os.makedirs(path)
```
willkg: TODO: Wrap this in a try/except and `logger.exception` anything that pops out.

willkg:
Yay! Circle is happy, so I'm happy. MERGING.

PR #57, opened 2016-09-29
Fix NoOpCrashStore logging

willkg:
Previously, if we got a raw_crash that was big with lots of keys, this
would log crazy unhelpful amounts of information. Now it logs
something more helpful.


willkg:
Circle is happy, so I'm going to land it.

PR #58, opened 2016-09-29
Move tests to tests/unittest/ with README

willkg:


willkg:
Looks good, so I'm landing it.

PR #59, opened 2016-09-29
Cleanup test Client

willkg:
This fixes some issues with the test Client where the client has a copy
of the app, but we need to rebuild the app with a certain configuration
for some tests and it was awkward to get the right order of operations.

Now you can just call .rebuild_app() and give it a new configuration
and you're all set.


willkg:
This looks good, so I'm going to land it.

PR #62, opened 2016-09-29
Implement AppConfig

willkg:
AppConfig lets us define application-scoped configuration that's global
to the whole project. We can pass bits of that as appropriate to
components that need it.

In this way, we can have good auto-generated documentation and less
tangling between components and the application.

### DIFF
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -26,6 +26,56 @@ logger = logging.getLogger(__name__)
 _logging_initialized = False
 
 
+class AppConfig(RequiredConfigMixin):
+    """Application-level config
+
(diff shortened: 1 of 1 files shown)


willkg:
I need to look into whether "config" is now being overloaded as a term.

willkg:
This looks good, so I'm going to land it.

Two notes:

1. I think "config" is overloaded as a term and variable name. I'll mull on how to rework things so it's easier.
2. The docstring for AppConfig probably isn't going to be useful for user-centric docs, so I might want to either nix it or tweak my sphinx extension to ignore the docstring. I forget if it can do that or not.

PR #65, opened 2016-09-30
Nix INI files

willkg:
All configuration will be done with environment variables going
forward.  At some point, it might behoove us to use env files--we can
do that then.


willkg:
Looks good! Merging!

PR #67, opened 2016-10-04
Breakpad util

willkg:
This implements `multipart_encode` which takes a Python dict of key/value pairs and encodes it in multipart/form-data for an HTTP POST.

I used this utility function to clean up the tests. We'll probably use it later for system tests, too.


willkg:
That fixes breakpad fs handling, too. It should be more robust despite `FieldStorage` parsing.

antenna/app.py
```diff
@@ -227,31 +192,31 @@ def extract_payload(self, req):
...
                 # crash that was re-submitted.
                 continue
 
-            elif isinstance(val, str):
-                raw_crash[key] = de_null(val)
+            elif fs_item.type.startswith('text/plain'):
+                raw_crash[fs_item.name] = de_null(fs_item.value)
 
-            elif isinstance(val, bytes):
+            elif fs_item.type.startswith('application/octet-stream') or isinstance(fs_item.value, bytes):
```
willkg: TODO: Verify that breakpad client actually puts content types in the part headers and that it's `text/plain` an `application/octet-stream`.

antenna/mini_poster.py
```diff
@@ -0,0 +1,69 @@
...
+def multipart_encode(raw_crash, boundary=None):
+    """Takes a raw_crash as a Python dict and converts to a multi-part/form-data
+
+    :arg params: Python dict of name -> value pairs. Values must be either
+         strings or a tuple of (filename, file-like objects with ``.read()``).
+
+    :arg boundary: The MIME boundary string to use. Otherwise this will be
+        generated.
+
+    :returns: bytes of a multi-part/form-data (which is quite large)
```
willkg: TODO: Fix this docstring. It returns a tuple of `(bytes, dict)` where the latter is headers with `Content-Type` and `Content-Length`.

willkg:
I'll fix those TODOs tomorrow. Looks good, though.

antenna/app.py
willkg: Breakpad client in Firefox doesn't send any `text/plain` content-type headers, but it does send `application/octet-stream`. I adjusted the code accordingly.

For example:

```
Content-Disposition: form-data; name="useragent_locale"

en-US
```

and:

```
Content-Disposition: form-data; name="upload_file_minidump"; filename="6da3499e-f6ae-22d6-1e1fdac8-16464a16.dmp"
Content-Type: application/octet-stream

<BYTES>
```

willkg:
This looks good, so I'm going to land it. It's possible I might tweak how these bits work later, but that's fine--iterative development ftw.

PR #68, opened 2016-10-04
Add load_raw_crash

willkg:
This will make it easier to freeze and thaw crashes from the various
crash storage systems. That makes it easier to use antenna as a
library to build utilities for testing and scripts.


willkg:
Looks good! MERGING!

PR #70, opened 2016-10-05
Rewrite ooid code

willkg:
Originally, I took the ooid code from socorrolib and then modified it
a bit. This redoes it so it's more convenient for our purposes.

Plus it changes the name from ooid to crash_id.


willkg:
Looks good. Merging!

PR #77, opened 2016-10-11
Fix docs creation

willkg:
This creates a bunch of docker stuff so that we can have a separate
docker image for doing development-specific things without having some
of the tools we need to install available in the deploy image.

The new local/antenna_dev image has make which we need to build the
docs.


willkg:
We can also split requirements so we're only installing the dev stuff in the dev container. I'll write up an issue for looking into that later.

willkg:
Circle is happy with it:

https://circleci.com/gh/mozilla/antenna/101

I talked to @pmac to see if the docker stuff was too convoluted. He says it's ok. It's similar to what we do with Standup, so there's that.

Merging it.

PR #80, opened 2016-10-12
Fix error handling

willkg:
Previously, unhandled errors would propagate all the way up and Falcon
would just return an HTTP 500 to the client, but wouldn't log anything
which ... is unhelpful.

This replaces that with a handler that logs the exception and then
returns a HTTP 500 to the client.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ build:
 	touch .docker-build
 
 run: .docker-build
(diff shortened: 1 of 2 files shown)


willkg:
Circle is happy and this is pretty straight-forward and if there are problems, we can fix them as we hit them. self-r+!

PR #88, opened 2016-10-17
Move post handling to the eventpool

willkg:
Prior to this change, we were saving the crash during the HTTP
request/response cycle.

This fixes it so that we spawn an eventlet to handle storing the crash
and whatever else needs to happen so that we can return a crash id to
the breakpad client and end the HTTP conversation as soon as possible.

We put all the eventlets in a pipeline_pool so that we can count them
and keep track of them.

This also moves some yucky code out of the tests and into the test
client to improve the readability of our test code.


willkg:
Looks super! MERGE!

PR #94, opened 2016-10-18
Implement AntennaAPI to manage resources

willkg:
Prior to this, falcon.API would register routes, but it was tricky to
get a list of resources. In Antenna, resources are also components, so
we want to have a list of the ones in the app so that we can more
easily perform work across all the components in the app.

For example, it's hard to do a health check if you can't access all
the components in the app to ask them about their health.

This fixes that.


willkg:
Looks good. Merging!

PR #96, opened 2016-10-18
Healthchecks

willkg:
Add `/__lbheartbeat__` and `/__heartbeat__` endpoints.


willkg:
Good enough for now. We can hone it later when/if other requirements surface. Merging!

PR #101, opened 2016-10-20
Fix NoneType issues

willkg:
If ANTENNA_ENV wasn't defined, then Antenna would raise an exception
on startup due to a bug in Everett.

This changes the code so it doesn't tickle the bug.


willkg:
No conflicts--merging!

PR #105, opened 2016-10-20
Switch around hacking environment

willkg:
This changes "make run" to kick off a prod-like environment rather
than a barebones one.

This also updates the README so it's helpful again.


willkg:
Looks good! LANDING!

PR #107, opened 2016-10-20
Support other s3 credentials options

willkg:
boto3 can figure out credentials via a variety of different methods.
We're planning to bind an IAM role to an EC2 instance, so we don't
need credentials at all.

This fixes the code so that it doesn't require ACCESS_KEY and
SECRET_ACCESS_KEY and adds a note in the docstring about it.

This doesn't add a test for that scenario. It's trickt to add a test
without reaching deep into the bowels of boto3 and mocking stuff out at
which point I'm not sure what we're actually testing. So... that's
stinky, but so it goes.


willkg:
Probably fine... We'll test it live!

PR #109, opened 2016-10-20
Fix error handling

willkg:
Previously, all "errors" were handled and then returned HTTP 500, but
that included things that should have been 404, 403, etc.

This fixes the error handling to do the right thing. All we wanted to do
that was different than what Falcon currently does is log unhandled
exceptions.


willkg:
Looks good. Yay!

PR #112, opened 2016-10-22
Fix logging and logging mock

willkg:
This adds a configurable logging level to setup_logging and the app
configuration.

This also fixes the logging mock and py.test fixture to work better
especially in regards to how setup_logging() only happens once.


willkg:
This is all set. Merging!

PR #113, opened 2016-10-22
Add a throttler

willkg:
This adds a throttler that behaves much like the throttler in the
current Socorro. There are some minor differences.

First, this throttler defines rule sets in Python. That makes things a
lot easier.

Second, this throttler merges DISCARD and IGNORE because in both cases
we're dropping the crash. Merging them simplifies the code a bit.

Third, this throttler names the rules. This will make logging better
plus it enables us to statsd our rules so we have some metrics about why
things tend to get accepted.

Fourth, other than configurable rule sets, the throttler is not
configurable. If you don't want throttling, use the "accept_all" rule
set.

### DIFF
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -21,6 +21,7 @@ from gevent.pool import Pool
 
 from antenna.datetimeutil import utc_now
 from antenna.util import create_crash_id, de_null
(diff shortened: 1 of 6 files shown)


willkg:
Looks good. Merging!

PR #117, opened 2016-10-27
Implement metrics infrastructure

willkg:
This implements infrastructure for metrics. It's global to the entire
app to make it easier to add metrics easily without having to add
additional tangling between things. The theory is that metrics is like
logging where it's available to the entire app because we always want
it.

This adds a LoggingMetrics and a DogStatsdMetrics. The former makes
development a little easier in some circumstances. The latter we'll
use in production.

This adds testing infrastructure to make it easy to verify specific
metrics are generated during execution of bits of code.


antenna/metrics.py
```diff
@@ -0,0 +1,332 @@
...
+                ...
+
+        """
+        def _inner(fun):
+            @wraps(fun)
+            def _timer_decorator(*args, **kwargs):
+                with self.timer(stat):
+                    return fun(*args, **kwargs)
+            return _timer_decorator
+        return _inner
```
willkg: This should get tested. I keep forgetting.

prod.env
```diff
@@ -5,6 +5,12 @@
 # DEBUG is helpful for development, but otherwise we'd use INFO
 LOGGING_LEVEL=DEBUG
 
+# Statsd things
+METRICS_CLASS=antenna.metrics.DogStatsdMetrics
+STATSD_HOST=statsd
+STATSD_PORT=8125
```
willkg: These two are the defaults, so we could nix them here.

tests/unittest/conftest.py
```diff
@@ -232,6 +240,27 @@ def _loggingmock(names=None):
...
+def metricsmock():
+    """Returns a metricsmock that builds a metrics mock context to record metrics records
+
+    Usage::
+
+        def test_something(metricsmock):
+            with metricsmock as mm:
+                # do stuff
+                assert mm.has_record(
+                    key='some.key',
```
willkg: Oops--that's wrong now. It's `stat`.

willkg:
That fixed a bunch of issues. I need to go through this again and double-check everything--I'll do that tomorrow.

antenna/metrics.py
```diff
@@ -0,0 +1,332 @@
...
+    :arg str extra: Any extra bits to add to the end of the key.
+
+    """
+    if type(thing) == type:
+        # It's a class, so the name is the module + class
+        thing = '%s.%s' % (thing.__module__, thing.__name__)
+    elif isinstance(thing, str):
+        pass
+    else:
+        thing = '%s.%s' % (thing.__class__.__module__, thing.__class__.__name__)
```
willkg: This feels awkward. Worth fixing.

antenna/metrics.py
```diff
@@ -0,0 +1,332 @@
...
+
+    return MetricsInterface(thing)
+
+
+class MetricsMock:
+    """Mock for recording metrics events and testing them"""
+    def __init__(self):
+        self.records = []
+        self._old_impl = None
+        self.client = self
```
willkg: We can nix this line--we're no longer calling into `.client` anymore.

antenna/metrics.py
willkg: This is tested now--we're using it in BreakpadSubmitterResource.

willkg:
Good enough. Landing.

PR #118, opened 2016-10-27
Add logging of exceptions

willkg:
When running in a worker process with gunicorn, it's possible the code
dies and then gunicorn shuts down, but nothing gets printed to the
console so you have no idea what happened.

This wraps .get_app() so it prints something to the console on
unhandled exceptions.

### DIFF
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -4,6 +4,7 @@
 
 import cgi
 from collections import OrderedDict
+from functools import wraps
 import hashlib
 import io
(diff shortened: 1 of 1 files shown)


willkg:
Landing this. Pretty sure it'll fix the startup problems I keep having.

willkg:
Where "fix" is more like "will tell me wtf is going on so I don't have to run gunicorn with `--preload` which causes other problems".

PR #119, opened 2016-10-28
Log config

willkg:


willkg:
It's kind of goofy to write infrastructure for logging configuration at startup, but ... so it goes.

PR #121, opened 2016-10-28
Fold datetimeutil into util

willkg:
There isn't any reason to have datetimeutil separate. This folds it in.


willkg:
Looks good!

PR #126, opened 2016-11-09
Cleanup

willkg:


willkg:
This is predominantly moving stuff around and then fixing docs. Merging.

PR #127, opened 2016-11-11
Bug #1316655: Upgrade Falcon to 1.1

willkg:
* upgrade Falcon to 1.1
* remove the test client I wrote for the new one in Falcon 1.1 which is
  much more py.test fixture friendly
* fix an issue with unbounded streams
* clean up startup exception handling


willkg:
Looks good. Merging!

PR #130, opened 2016-11-29
Bug 1316431: Add /__broken__ endpoint

willkg:
This adds a /__broken__ endpoint which we can use to trigger exceptions
which Sentry will catch which lets us test Sentry is working.


willkg:
Looks good.

PR #131, opened 2016-12-10
System test

willkg:
This cleans up some test relevant stuff I threw together and then had scattered across the repository. It also pulls in the system test stuff I wrote in another repo called jansky and adds some scaffolding for conveniently running the tests using environment files.

@milescrabill and i still have to figure out how these will get run in CI--that's not hooked up, yet.


willkg:
I have no idea why Circle isn't running the tests. However, there's nothing in here that would be affected by unit tests, so it's sort of moot.

Future me will self-review this in a couple of days.

mcrabill-mozilla-readonly:
It might also have been that this account, my "read only" account, did not have admin on the repo. It now does. This was the account that the hook was added by.

PR #133, opened 2016-12-14
Bug 1323462: Remove extra / for s3 paths

willkg:
This removes the extra / for s3 paths so they don't end up in a
//v2/raw_crash/... kind of pseudo-filename.


willkg:
Landing this. Then I'll test it on -dev and verify it's correct.

PR #136, opened 2016-12-14
Fix content-length tests

willkg:
This cleans up the content-length handling code and the tests--the
test_no_content_length test was accidentally sending a content length.

This adds a test for when the content-length is more than the size of
the payload. This requires testing against nginx which will timeout the
connection after a minute (the right thing to do).


willkg:
Looks good. Merging.

PR #137, opened 2016-12-14
Bug 1322077: Fix zlib issue

willkg:
If the payload isn't compressed or zlib throws an error when
uncompressing the payload, we should treat it as junk and move on.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -146,7 +146,14 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
             # have to do that here because nginx doesn't have a good way to do
             # that in nginx-land.
             gzip_header = 16 + zlib.MAX_WBITS
-            data = zlib.decompress(req.stream.read(content_length), gzip_header)
+            try:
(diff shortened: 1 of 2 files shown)


antenna/breakpad_resource.py
```diff
@@ -146,7 +146,14 @@ def extract_payload(self, req):
...
             gzip_header = 16 + zlib.MAX_WBITS
-            data = zlib.decompress(req.stream.read(content_length), gzip_header)
+            try:
+                data = zlib.decompress(req.stream.read(content_length), gzip_header)
+            except zlib.error:
+                # This indicates this isn't a valid compressed stream. Given
+                # that the HTTP request insists it is, we're just going to
+                # assume it's junk and not try to process any further.
+                self.mymetrics.incr('bad_gzipped_crash')
+                return {}, {}
```
willkg: We don't want to bother logging these. We do want to keep tabs on whether this happens a lot--my current theory is that it rarely happens if ever. If this number is higher than we want, we can add logging and look into it then.

willkg:
Looks good. Merging.

PR #138, opened 2016-12-14
Bug 1323464: Implement system test script

willkg:
* implements a system test script so this is easier to run in a docker
  container in the -dev environment
* renames prod.env to dev.env which is more appropriate
* updates documentation
* cleans up some oddities


willkg:
@milescrabill After this, you can spin up a docker container with a `.env` file and run `./tests/systemtest/run_tests.sh` and it'll do all the work it needs to do.

Does that look ok?

willkg:
Looks good to me. I'll land it and talk with @milescrabill when he has a free moment.

PR #141, opened 2017-01-09
1329736 dumps

willkg:


willkg:
I'll self-review this tomorrow.

antenna/ext/crashstorage_base.py
```diff
@@ -17,14 +17,14 @@ class CrashStorageBase(RequiredConfigMixin):
     def __init__(self, config):
         self.config = config.with_options(self)
 
-    def save_raw_crash(self, raw_crash, dumps, crash_id):
+    def save_raw_crash(self, crash_id, raw_crash, dumps):
```
adngdb: :+1: 

This is a change I proposed in https://github.com/adngdb/crashmill. I also proposed that we separate ``save_raw_crash`` and ``save_dumps``, since the two actions are distinct anyway, and it made the API a little more coherent in my opinion. Cf. https://github.com/adngdb/crashmill/blob/master/crashmill/crash_store.py

willkg: On the call, I was thinking "seems better to save everything in one pass rather than split it up", but now I'm thinking that's a lot of extra work if you're just making changes to the raw crash metadata and want to save that.

I'm persuaded. I'll split it up into `save_raw_crash` and something else. Maybe `save_dumps` or maybe I'll just do `save_dump` and move the iteration to the parent.

willkg:
Looks good. I'll do the `save_raw_crash` changes in another PR.

PR #142, opened 2017-01-10
Bug 1329998: Split save_raw_crash up

willkg:
This splits save_raw_crash into save_raw_crash and save_dumps. The reason for
this is so that we can save the raw crash independently of the dumps if we need
to and thus reduce S3 work.


willkg:
I wanted to do `save_dump`, but we also need to save the `dump_names` file, so it helps to save all the dumps at once. Given that we don't edit dumps (at least, I've never seen any evidence we do), I think this is ok.

willkg:
I'll review this tomorrow.

willkg:
Looks good to me. Merging.

PR #143, opened 2017-01-13
Fix content-length systemtests to correctly use https

willkg:
If the POSTURL is an https url, then the content-length tests would fail because
we were always using HTTPConnection. This fixes that.

@milescrabill Pretty sure this fixes it, but I can't test it locally, so ... meh. Does it look ok?


willkg:
This doesn't get run in the Circle CI tests, so whether they pass or not is moot.

willkg:
@milescrabill said this looked fine. Merging it.

PR #145, opened 2017-01-13
Fix content-length test and post-crash tests

willkg:
* the content-length test should be checking for a 504--not a 200
* the post-crash tests were verifying using pseudo-filenames that ended up with
  // in them which I'm pretty sure caused them to fail
* added some more debug logging so failures are easier to debug

@milescrabill Does this look ok?

### DIFF
--- a/tests/systemtest/test_post_crash.py
+++ b/tests/systemtest/test_post_crash.py
@@ -27,14 +27,14 @@ class CrashVerifier:
         self.errors.append(msg)
 
     def raw_crash_key(self, crash_id):
-        return '/v2/raw_crash/{entropy}/{date}/{crashid}'.format(
+        return 'v2/raw_crash/{entropy}/{date}/{crashid}'.format(
             entropy=crash_id[0:3],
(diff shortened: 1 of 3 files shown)


tests/systemtest/test_post_crash.py
```diff
@@ -27,22 +27,22 @@ def add_error(self, msg):
         self.errors.append(msg)
 
     def raw_crash_key(self, crash_id):
-        return '/v2/raw_crash/{entropy}/{date}/{crashid}'.format(
+        return 'v2/raw_crash/{entropy}/{date}/{crashid}'.format(
```
willkg: The `/` at the beginning would cause the full pseudo-filename to have a `//` in it which is a different path than what Antenna is saving to.

I fixed this issue in Antenna in https://github.com/mozilla/antenna/commit/7c6637a826c4e2b7ab6adff80d776f9ecee71863 but must not have carried it through to the system tests. My bad. :(

willkg:
Thank you!

PR #146, opened 2017-01-17
Improve mini_poster cli

willkg:
This lets the cli post crashes given a raw_crash and zero or more dumps. You can
use this to assemble crashes using arbitrary bits.


willkg:
This is a helpful tool for posting test crashes to antenna and socorro's collector and comparing results.

PR #148, opened 2017-01-30
Improve mini_poster

willkg:
* improve logging so it's more informative
* add "mode" where if you give it a raw crash that's obviously in a directory
  structure that's similar to how we save things on s3, it will find the other
  parts and assemble the crash accordingly


willkg:
Looks good--merging.

PR #153, opened 2017-02-03
Bug 1336533: Switch gauge to histogram

willkg:
We want histogram stats for crash sizes since we want to know median, 95% and
max. Gauge is the wrong thing for that. This fixes it.


willkg:
Looks good, I think.

PR #154, opened 2017-02-07
Move Dockerfile gunicorn config to env vars

milescrabill:
@willkg r?


willkg:
Looks good as far as I can tell. r+!

PR #155, opened 2017-02-08
Change user/group ID to prevent overlap

milescrabill:
1001 corresponds to a Puppet-added user. We should change that.

@willkg r?


willkg:
This PR had me in tears. The ending was totally unexpected. Well done! r+

PR #156, opened 2017-02-08
Bug 1337859: Capture end-to-end timing

willkg:
This adds a metric that tells us the total time it took between getting the HTTP
POST request to saving all the bits to s3. We do it as a timing so we can get
median, 95% and max rates in Datadog.


willkg:
Looks good. MERGING!

PR #157, opened 2017-02-08
Add test with large payload and wrong content-length

willkg:
We want to make sure that nginx is rejecting large payloads based on the actual
size of the payload and not the content-length.

Note: I can't test this locally at the moment because we're not using nginx in local development environments. So I'm going to land this and then "test" it when the systemtest runs for -dev.


willkg:
I think this looks ok, so I'm mergifying it.

m8ttyB:
r+ smells good

PR #158, opened 2017-02-09
Update requirements to latest

willkg:


willkg:
This seems fine. I ran systemtests locally, though that doesn't run them all. We'll know more when this gets to -dev.

PR #160, opened 2017-02-10
pip install newrelic in the Dockerfile

milescrabill:
@willkg r?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,8 @@ RUN apt-get update && \
 COPY ./requirements.txt /app/requirements.txt
 
 RUN pip install -U 'pip>=8' && \
-    pip install --no-cache-dir -r requirements.txt
+    pip install --no-cache-dir -r requirements.txt\
+    pip install newrelic
 
 # Install the app
 COPY . /app/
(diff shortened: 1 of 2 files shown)


milescrabill:
Interestingly, it cannot find newrelic. It is in pypi and installs just fine on a local Python 3.5.2. @willkg halp I am not Pythonista.

Dockerfile
```diff
@@ -9,7 +9,8 @@ RUN apt-get update && \
 COPY ./requirements.txt /app/requirements.txt
 
 RUN pip install -U 'pip>=8' && \
-    pip install --no-cache-dir -r requirements.txt
+    pip install --no-cache-dir -r requirements.txt\
+    pip install newrelic
```
willkg: This is missing an `&&`, so it's concatenating this line and the next as a single "pip install" and there's no package named "install".

Further, we really want to pin newrelic and use a hash, so I'm going to move that to a `requirements_newrelic.txt` file.

willkg:
^^^ That should fix the issue plus it pins the package. I'll wait for Travis to be ok, then merge it.

willkg:
Now that I'm thinking about this, since it's in the `Dockerfile` it gets installed for everyone anyhow, so we might as well just put it in the `requirements.txt` file with everything else.

Any objection to that? If not, I can make that change.

willkg:
I talked with @milescrabill and he said we should keep it as two separate requirements files for now. Given that, this looks super! r+

PR #161, opened 2017-02-10
Add an env var for command prefix to Dockerfile

milescrabill:
@willkg r?

This is sort of a different solution, we could also have a conditional bash script that runs Antenna, but this way I can just set $CMD_PREFIX with Puppet and things are cool.


willkg:
Looks good to me. r+

PR #162, opened 2017-02-14
Bug 1339403: Fix TypeError

willkg:
Fixes a bug where the wrong argument type was being passed into .add_error().


willkg:
I can't write a test for this without mocking a lot of boto3's internals and that seems not worth it. The specific problem is pretty obvious (passing in a non-str argument to a method expecting a str), so I'm just going to fix it and move on.

self-r+ and merging.

PR #163, opened 2017-02-24
Bug 1342525: Fix queuing of s3 uploads

willkg:
Previously, uploading crash data to s3 would happen concurrently and Antenna
would spin off a new coroutine for every crash it was handling. That doesn't
scale.

This changes it so that there's an explicit queue and we cap the number of
actively saving-to-s3 coroutines at CONCURRENT_SAVES.


willkg:
This looks good. I'm going to land it, wait until it hits dev and then see how it goes. Then we should deploy to stage.

PR #164, opened 2017-02-27
Bug 1342619: Make boto3 client creation retryable

willkg:
This reworks the retry code so it's easier to work with. It also changes boto3
client creation so that it gets retried a few times if it fails.

The theory here is that in Antenna -stage, the boto3 lib does an HTTP request to
get the credentials, gets rate-limited and then fails and that hoses Antenna's
ability to start up. Assuming the theory is right, this will alleviate that a
bit.

### DIFF
--- a/antenna/ext/s3/connection.py
+++ b/antenna/ext/s3/connection.py
@@ -17,6 +17,17 @@ from antenna.util import retry
 logger = logging.getLogger(__name__)
 
 
+def wait_times_save():
+    """Wait time generator for failed save attempts
+
+    This waits 2 seconds between failed save attempts for 5 iterations and then
+    gives up.
+
+    """
+    for i in [2, 2, 2, 2, 2]:
+        yield i
+
+
 class S3Connection(RequiredConfigMixin):
     """Connection object for S3.
 
@@ -94,6 +105,16 @@ class S3Connection(RequiredConfigMixin):
             # anything with them.
             self.verify_configuration()
 
+    @retry(
+        retryable_exceptions=[
+            # FIXME(willkg): Seems like botocore always raises ClientError
+            # which is unhelpful for granularity purposes.
+            ClientError,
+        ],
+        wait_time_generator=wait_times_save,
+        sleep_function=gevent.sleep,
+        module_logger=logger,
+    )
     def _build_client(self):
         # Either they provided ACCESS_KEY and SECRET_ACCESS_KEY in which case
         # we use those, or they didn't in which case boto3 pulls credentials
@@ -140,10 +161,25 @@ class S3Connection(RequiredConfigMixin):
             # which is unhelpful for granularity purposes.
             ClientError,
         ],
+        wait_time_generator=wait_times_save,
         sleep_function=gevent.sleep,
         module_logger=logger,
     )
     def save_file(self, path, data):
+        """Saves a single file to s3
+
+        This will retry a handful of times in short succession so as to deal
+        with # some amount of fishiness. After that, the caller should retry
+        saving after a longer period of time.
+
+        :arg str path: the path to save to
+
+        :arg str data: the data to save
+
+        :raises botocore.exceptions.ClientError: connection issues, permissions
+            issues, bucket is missing, etc.
+
+        """
         if not isinstance(data, bytes):
             raise TypeError('data argument must be bytes')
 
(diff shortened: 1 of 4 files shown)


antenna/ext/s3/connection.py
```diff
@@ -17,6 +17,17 @@
...
 
+def wait_times_save():
+    """Wait time generator for failed save attempts
+
+    This waits 2 seconds between failed save attempts for 5 iterations and then
+    gives up.
+
+    """
+    for i in [2, 2, 2, 2, 2]:
+        yield i
```
willkg: This is a much nicer construct for dealing with "what's the next wait time" as well as "have I exceeded the maximum number of retry attempts". Previously, the retry attempts were hard-coded and we'd pass in a `max_attempts` argument. That felt like a clumsier API.

antenna/util.py
```diff
@@ -139,21 +139,17 @@ def __init__(self, msg, ret):
...
-)
-RETRY_TIMES_MAX_INDEX = len(RETRY_TIMES) - 1
-
-
-def retry(retryable_exceptions=Exception, retryable_return=None, max_attempts=10,
-          sleep_function=time.sleep, module_logger=None):
-    """Decorator for retrying with exponential wait and logging
+def wait_time_generator():
+    for amt in [1, 1, 5, 10, 30]:
+        yield amt
```
willkg: This is a much nicer construct for dealing with "what's the next wait time" as well as "have I exceeded the maximum number of retry attempts". Previously, the retry attempts were hard-coded and we'd pass in a `max_attempts` argument. That felt like a clumsier API.

antenna/ext/s3/connection.py
```diff
@@ -140,10 +161,25 @@ def check_health(self, state):
...
+        This will retry a handful of times in short succession so as to deal
+        with # some amount of fishiness. After that, the caller should retry
+        saving after a longer period of time.
+
+        :arg str path: the path to save to
+
+        :arg str data: the data to save
+
+        :raises botocore.exceptions.ClientError: connection issues, permissions
+            issues, bucket is missing, etc.
```
willkg: This (and other docstring fixes in this PR) codifies who's in charge of dealing with what exceptions. That's important for understanding retryable things and their effects on other things.

antenna/util.py
```diff
@@ -211,25 +206,51 @@ def _retry_inner(fun):
...
+            wait_times = wait_time_generator()
+
             while True:
+                # Grab the next wait time and if there isn't one, then this is
+                # the last attempt
+                try:
+                    next_wait = next(wait_times)
+                except StopIteration:
+                    next_wait = None
+
```
willkg: So, this function got complicated by the fact that it does different things based on whether the last attempt threw an exception or returned a "failed result" coupled with the fact that when iterating over a generator, you will never know if the next iteration will halt.

To deal with that, for each iteration, we pull the next wait time before doing the iteration and in this way we know whether this is the last one or not. That's why this isn't in a `for next_wait in wait_time_generator():` kind of loop.

antenna/util.py
```diff
@@ -211,25 +206,51 @@ def _retry_inner(fun):
...
+                    # If last attempt, then raise MaxAttemptsError
+                    if next_wait is None:
+                        raise MaxAttemptsError('Maximum retry attempts.', ret)
+
+                except retryable_exceptions as exc:
+                    # Retryable exception is thrown, so we log something and
+                    # then do another iteration.
                     if module_logger is not None:
-                        module_logger.exception('retry attempt %s', attempts)
+                        module_logger.warning(
```
willkg: I ditched `.exception()` because that's pretty spammy. If the last attempt fails, it can print the entire traceback. I switched it to `.warning()` figuring that this shouldn't happen, but it's getting retried, so this particular instance of it isn't an "error" that needs to be dealt with.

antenna/util.py
```diff
@@ -211,25 +206,51 @@ def _retry_inner(fun):
...
-                    elif max_attempts is not None and attempts >= max_attempts:
-                        raise MaxAttemptsError(
-                            'Maximum retry attempts. Error return.', ret
+                    # The return value is "bad", so we log something and then
+                    # do another iteration.
+                    if module_logger is not None:
+                        module_logger.warning(
+                            '%s: bad return, retry attempt %s',
+                            fun.__qualname__,
+                            attempts
```
willkg: I added logging here because it occurred to me we'd like some heads-up when things are going wonky regarding returned results.

willkg:
I think this looks ok. I'm going to merge it.

PR #165, opened 2017-02-27
Bug 1342583: Capture startup errors in sentry

willkg:
Currently, if Antenna is sad or despondent at startup, we are blissfully
unaware. That turns out to be problematic. We want to know if our little Antenna
is having problems getting out of bed in the morning.

This fixes the startup code to send startup errors to Sentry.

### DIFF
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -244,35 +244,45 @@ def get_app(config=None):
             )
 
         app_config = AppConfig(config)
-        setup_logging(app_config('logging_level'))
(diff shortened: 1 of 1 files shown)


willkg:
This diff looks complicated, but it's not very interesting. I moved creation of the Sentry client higher in the code, moved everything into a bit `try/except` and added a `sentry_client.captureException()` call. The rest is pretty much the same.

Given that, merging.

PR #168, opened 2017-02-27
Bug 1342606: Move health stats to a heartbeat

willkg:
Previously, Antenna would emit health stats for the save queue size and active
worker pool size as statsd data when /__heartbeat__ was handled. However, that
only gets handled by one process on the node, so the data we're getting back
from a node isn't the whole story and we won't get back data from orphaned
processes that aren't handling HTTP requests.

This changes the architecture so that BreakpadResource has a heartbeat coroutine
that sleeps 30 seconds and then emits stats.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -13,6 +13,7 @@ import zlib
 from everett.component import ConfigOptions, RequiredConfigMixin
 from everett.manager import parse_class
 import falcon
+import gevent
 from gevent.pool import Pool
 
 from antenna import metrics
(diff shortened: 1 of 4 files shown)


antenna/breakpad_resource.py
```diff
@@ -140,14 +144,28 @@ def get_runtime_config(self, namespace=None):
...
+            self.crashstorage.check_health(state)
+
+    def heartbeat(self):
+        """Heartbeat function
+
+        Every 30 seconds, runs a set of methods on this class.
+
+        """
+        while True:
+            gevent.sleep(30)
```
willkg: Sleep first, then do stuff because there's no need to do stuff immediately.

willkg:
Looks fine. I talked with @milescrabill and he said 30s is a good number, so that's what we're going with. If we need to make it configurable later, that's cool, but for now 30s probably fine.

antenna/breakpad_resource.py
lonnen: Could you drop this in the code plz?

PR #169, opened 2017-02-28
Fix logging and metrics

willkg:


willkg:
Logging happens out of order which is confusing when reading the logs. This reduces the number of lines logged and thus should reduce some confusion.

This also tweaks the `crash_save` `.incr` and changes the name so we're not cursed with bad data.

PR #173, opened 2017-03-02
Bug 1343708 batch metrics

willkg:


antenna/heartbeat.py
```diff
@@ -0,0 +1,92 @@
...
+
+    This holds heartbeat state and the methods used to start, stop and run the
+    heartbeat.
+
+    """
+
+    required_config = ConfigOptions()
+
+    # Interval between heartbeats
+    heartbeat_interval = 10
```
willkg: Why 10? Seemed like a good number that's long enough that batching is helpful but not so long that it's annoying.

antenna/heartbeat.py
```diff
@@ -0,0 +1,92 @@
...
+
+class HeartbeatManager(RequiredConfigMixin):
+    """Heartbeat manager
+
+    This holds heartbeat state and the methods used to start, stop and run the
+    heartbeat.
+
+    """
+
+    required_config = ConfigOptions()
```
willkg: It doesn't use any configuration options currently, but it's easier if components all behave the same.

antenna/heartbeat.py
```diff
@@ -0,0 +1,92 @@
...
+
+        """
+        while self.hb_live:
+            logger.debug('thump')
+            for fun in _registered_hb_funs:
+                try:
+                    with capture_unhandled_exceptions():
+                        logger.debug('hb: running %s', fun.__qualname__)
+                        fun()
+                except Exception:
```
willkg: We want to allow for vm problems to just kill this off particularly `KeyboardInterrupt` and `SystemExit`.

https://docs.python.org/3.5/library/exceptions.html#exception-hierarchy

We also want to wrap each individual function so that it's not the case that one function fails and then prevents other functions from running.

We want to capture all unhandled exceptions so they show up in Sentry so we know there are things to fix.

antenna/metrics.py
```diff
@@ -183,13 +200,28 @@ def __init__(self, name):
...
+
+        This is the same as ``.incr()``, however it gets batched and sent out
+        every heartbeat_interval.
+
+        This is helpful for counters that count more frequently than 1/s where
+        statsd might "do interesting things" such that you don't get good
+        counts.
+
+        """
+        _metrics_impl.incr(self._full_stat(stat), value=value)
```
willkg: We could do the batching/flushing in this class, but ... there's possibly one of these for every component in Antenna. Instead, I decided to do it in the implementation since there's only one of those at any given time.

willkg:
This feels good. I'm going to land it.

PR #174, opened 2017-03-02
Redo crashmoving

willkg:
This renames some things to "crashmover" and related names. This should make it
easier to understand in the context of the current Socorro collector that this
replaces.

This changes the number of concurrent crashmovers running. Previously, we had 10
running. But it was actually (10 * number-of-processes) which is 50 per node and
they were all fighting for upload bandwidth. This changes that to 1 so there are
at most 5 per node. That should work a *lot* better.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -116,7 +116,7 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
 
     # Maximum number of concurrent crashmover workers; each process gets this
     # many concurrent crashmovers, so if you're running 5 processes on the node
-    # then it's (5 * concurrenct_crashmovers) fighting for upload bandwidth
+    # then it's (5 * concurrent_crashmovers) fighting for upload bandwidth
     concurrent_crashmovers = 1
 
     def __init__(self, config):
@@ -133,7 +133,7 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
         self.mymetrics = metrics.get_metrics(self)
 
         # Register hb functions with heartbeat manager
-        register_for_heartbeat(self.hb_health_stats)
+        register_for_heartbeat(self.hb_report_health_stats)
 
     def get_runtime_config(self, namespace=None):
         for item in super().get_runtime_config():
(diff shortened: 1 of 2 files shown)


antenna/breakpad_resource.py
```diff
@@ -113,22 +113,23 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
...
-    required_config.add_option(
-        'concurrent_saves',
-        default='10',
-        parser=positive_int,
-        doc='max number of crash reports being saved concurrently; minimum of 1'
-    )
+
+    # Maximum number of concurrent crashmover workers; each process gets this
+    # many concurrent crashmovers, so if you're running 5 processes on the node
+    # then it's (5 * concurrenct_crashmovers) fighting for upload bandwidth
```
willkg: Oops--typo in "concurrent".

antenna/breakpad_resource.py
```diff
@@ -150,10 +151,10 @@ def check_health(self, state):
 
     def hb_health_stats(self):
         # The number of crash reports sitting in the queue
```
willkg: It's worth mentioning that this is a key "health of this process" metric.

antenna/breakpad_resource.py
```diff
@@ -150,10 +151,10 @@ def check_health(self, state):
...
-        self.mymetrics.gauge('save_queue_size', len(self.save_queue))
+        self.mymetrics.gauge('save_queue_size', len(self.crashmover_save_queue))
 
         # The number of actively running coroutines saving crashes
-        self.mymetrics.gauge('active_save_workers', len(self.pipeline_pool))
+        self.mymetrics.gauge('active_save_workers', len(self.crashmover_pool))
```
willkg: Given that this is either 0 or 1, I'm not sure we need this anymore. But since it's sent during heartbeats, might as well keep it.

willkg: We don't need this anymore. It's either 0 or 1 and most of the time when measured it'll probably be 0. So ... that doesn't tell us anything useful.

willkg:
Looks good. Landing.

PR #176, opened 2017-03-08
ci: also deploy antenna container to dockerhub with circle build num

milescrabill:
This will help correlate builds across our many building things

@willkg r?


willkg:
Seems fine to me. We'd know if it was bad pretty quick.

PR #177, opened 2017-03-09
Adjust crashmover spawning

willkg:
Previously, a crashmover coroutine would spawn after an on_post. However, for
hand-wavey reasons, maybe a coroutine could "die" and if there's not an on_post
right afterwards, another one won't get spawned and Antenna might sit around for
a bit and not save something.

Anyhow, this adjusts things so that the heartbeat also kicks off coroutines if
appropriate.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -48,6 +48,9 @@ class SaveQueue:
     def __init__(self):
         self._queue = []
 
+    def __bool__(self):
+        return bool(len(self))
+
     def __len__(self):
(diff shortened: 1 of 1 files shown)


willkg:
This tries to kick off crashmover coroutines more often. It may have no net effect. The whole "coroutines might go away" theory is super hand-wavey.

However, this does give us a "safety guarantee' that every 10 seconds it will check to see if any are running and kick off a new one if there needs to be.

willkg:
This seems good.

PR #180, opened 2017-03-13
Hb in pool

willkg:


antenna/heartbeat.py
```diff
@@ -63,7 +62,9 @@ def heartbeat(self):
         capture unhandled exceptions and report them.
 
         """
-        while self.hb_live:
+        # Keep beating unless the WSGI worker is shutting down AND all
+        # registered lifers are ok with us shutting down
+        while self.is_alive() or any([fun() for fun in _registered_lifers]):
```
willkg: The `or` here is short-circuiting, so if `is_alive()` returns `True`, then it doesn't call any of the lifers.

bin/run_web.sh
```diff
@@ -21,9 +21,10 @@ gunicorn \
     --reload \
     --bind=0.0.0.0:8000 \
     --workers=1 \
-    --worker-connections=1 \
+    --worker-connections=2 \
```
willkg: This number now covers incoming connections AND the heartbeat because they're in the same pool. Thus if it's 1, then Antenna hilariously doesn't work at all because the heartbeat takes the slot and keeps it forever.

PR #181, opened 2017-03-14
Bug 1347185: fix s3 connection problems

willkg:
This changes the code to retry ValueError when trying to build a client. We get
a ValueError when boto3 uses the wrong endpoint.

Our theory is that it tries to get credentials and fails for some reason, then
degrades to this endpoint and fails.

Now if it hits this scenario (which seems to happen a lot when we start a new
node), it'll retry 5 times over 10 seconds (might not be enough--we'll find
out).

This also logs a line if it sent something to Sentry.

### DIFF
--- a/antenna/ext/s3/connection.py
+++ b/antenna/ext/s3/connection.py
@@ -18,7 +18,7 @@ logger = logging.getLogger(__name__)
 
 
 def wait_times_save():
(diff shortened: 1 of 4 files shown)


willkg:
After you merge, make sure to watch the grephost logs on -dev to see it deploy and what happens at startup and whether this made anything better.

PR #191, opened 2017-03-27
Bug 1349995: Allow changing host id in logs

willkg:
This makes it possible to set the host id in the logs to whatever you want using
the HOST_ID environment variable. This makes it easier to set the host id to
something that you can later use to correlate items in logs to other data
systems.


willkg:
Looks good.

PR #192, opened 2017-04-01
Fix create_s3_bucket.py script and add it to run_web.sh

willkg:
This fixes the script a bit so that it's more informative in what it's doing.
Further, this adjusts run_web.sh so that it runs the script creating the bucket
if necessary. This reduces a bunch of first-time developer pain.


willkg:
This shouldn't affect -stage and -prod environments at all. Should only affect local development.

PR #193, opened 2017-04-03
Send commit as tag to Sentry

willkg:
This adjusts the Sentry client creation so that it sends the commit sha of the
deployed code as a tag with all Sentry reports.

The reason we do this instead of sending the release is that this doesn't
require us to create the release every time we do a deploy.


willkg:
Looks good!

PR #197, opened 2017-04-05
Docs pass

willkg:


willkg:
Good enough for a first pass. We can improve from here.

PR #198, opened 2017-04-06
Fix S3 Connection to align better with boto3 creds usage

willkg:
This fixes the root: vars for the S3 Connection and the documentation so it
aligns better with boto3's credentials usage. Why do this? It reduces some
unneeded complexity.


willkg:
Looks good. Merging.

PR #202, opened 2017-04-09
Bug 1343708: Remove batch_incr

willkg:
We were having problems where graphs weren't lining up. I did some work and
theorized that since Antenna sends the same value a lot, that some of them were
getting dropped during the "normalize for a flush interval" period. Thus I
implemented batching of counters that occur often.

Since then, I've learned a lot more about statsd and I think all those theories
are rubbish.

Given that the theories are rubbish and while in the bug comments I claim that
batching helped, I don't think it actually did help at all and I think I was
interpreting the data noise incorrectly.

This commit removes the batching code because it's probably not actually helping
at all and thus it's adding unneeded code complexity.


willkg:
Yay!

PR #204, opened 2017-04-14
Add arg types to args in docstrings

### LINKED ISSUE #75: add arg types to docstrings
I want to go through and add argument types to docstrings. It makes some things easier.

For example, instead of:

```
    :arg foo: some foo thing
```

we do:

```
    :arg int foo: some foo thing
```

It shows up nicely in the Sphinx docs.

willkg:
I had done a bunch of args already. This fixes the remaining ones.

Fixes #75.


willkg:
This just affects the docs--it's cosmetic.

PR #205, opened 2017-04-14
Bug 1337506: Shutdown fix

willkg:
This fixes shutdown so it works correctly. Amongst other things, you can set `max_requests` and watch Antenna serve x requests, make sure everything got saved, and then get recycled by Gunicorn. It's super.


antenna/gunicornhooks.py
```diff
@@ -2,65 +2,36 @@
...
-    # but every method I found is a gross trade-off. Since the value of
-    # worker_connections is 1000, this seems pretty safe.
-    for obj in gc.get_objects():
-        if isinstance(obj, Pool):
-            if obj.size == worker_connections:
-                pool = obj
-                break
-    else:
-        print('No Pool instantiated with %s slots.' % worker_connections)
-        return
```
willkg: Thank goodness that is gone now.

willkg:
Looks so much better than what it's replacing. Yay!

PR #208, opened 2017-04-15
Bug 1356671: Fix crash saving exception handling

willkg:
* change crash saving retrying so that it only retries 20 times and then gives
  up; each time it sends something to sentry so we shouldn't miss it.
* add a test for the new retry behavior


willkg:
Looks good.

PR #209, opened 2017-04-15
Bug 1356671: Blind fix for None keys in raw_crash

willkg:
I can't for the life of me figure out how a None can end up as a key in the
raw_crash. Everything I do to try to coerce extract_payload() into yielding a
None is for naught. Thus this is a blind fix for something I can't reproduce.


willkg:
One important thing about this fix is that it can't make the situation worse or cause a different bug. It's just makes the existing code more defensive.

PR #212, opened 2017-04-17
Bug 1351914: Switch to moto for fake s3

willkg:
This switches from fake s3 to moto. This solves several problems the biggest of
which being that they no longer distribute fakes3 under and Open Source license.

The other problem this solves is that fakes3 doesn't let you list buckets which
we now need.

This uses a specific commit in the master branch after release 0.4.31 because
that release has bugs that prevent us from creating buckets.


willkg:
This affects development--it doesn't affect production in any way. Any goofs this introduces we can fix along the way.

PR #213, opened 2017-04-17
Bug 1356319: Simplify systemtest and document AWS S3 permission requi…

willkg:
…rements


willkg:
Looks good. This is a change to the system test, so we'll know whether it's good or not for sure after it runs on -dev. I'll fix anything that's busted in a future PR.

PR #217, opened 2017-04-18
Bug 1356899: Upgrade to python 3.6.1

willkg:
This upgrades the Docker image we base all our stuff on from
python:3.5.1-slim (which is no longer available) to python:3.6.1-slim.

I also updated systemtest requirements since the boto3 version it was using
didn't support Python 3.6.


willkg:
@milescrabill The python:3.5.1-slim we use now is gone. So we could update to python:3.5.2-slim and that's virtually the same ... or we could switch to python:3.6.1-slim which involves some library changes plus a more compact dict implementation (and some other things).

The unit tests all pass with 3.6. The system test works locally. I think we'd know quickly if this was ok on -dev if I threw some crashes at it. I suspect this is the kind of upgrade where we'd know one way or the other pretty fast--our tests cover most of the code already.

Should we upgrade to 3.6.1 now or upgrade to 3.5.2 now and wait for a few weeks on 3.6.1?

willkg:
I talked to @milescrabill and he said that every time he thinks of Python 3.6.1, he thinks of the smell of crabapple trees in spring.

I'm going to update this PR, then land it, and push it to -stage. We'll let it sit for a bit, then push it to -prod.

PR #223, opened 2017-04-21
Bug 1357944: Move incr for throttle rule

willkg:
This moves the incr for the throttle rule into the breakpad_resource to a place
where we have the rule used even if we picked a rule without running through the
throttler.

This will additionally cover NO_MATCH, THROTTLEABLE_0, ALREADY_THROTTLED, and
FROM_CRASHID.


willkg:
I missed this in the previous pass. We want to know how often these "fake rules" happen as well as when the actual rules happen, so moving the `mymetrics.incr` call is a good idea.

PR #231, opened 2017-04-28
Switch to localstack

willkg:
This switches from the moto container I was using to localstack which is
essentially the same thing, but maintained a little better and I don't have to
maintain my own Dockerfile.


willkg:
This only affects local development. It's a better situation than moto since we don't have to maintain our own Dockerfile. Plus with other Socorro spin-off projects, this might be helpful since it supports other AWS services.

Given all that, I'm going to merge it.

PR #239, opened 2017-05-16
Ditch virtualenv requirements

willkg:
Python 3.3+ has virtualenv built-in. This drops the library for the built-in one.


willkg:
Looks super!

PR #241, opened 2017-05-18
Bug 1365416: Fix throttle character checking

willkg:
This fixes the throttle character checking code to no longer assume it's an
int-like thing.

Further, it reworks all the tests for getting the throttle result and cleans
them up significantly.


willkg:
Looks good. Merging!

PR #242, opened 2017-05-31
circleci conf: only do things if DOCKER_DEPLOY=true

milescrabill:
@willkg r?


willkg:
Ha!

This looks fine to me as long as it works. I'm not really sure how to determine whether it works without landing it.

willkg:
I'll just land it and we can see what happens tomorrow.

PR #243, opened 2017-06-22
Bandit coverage

willkg:


willkg:
This is mostly cosmetic, but also adding the bandit linter. Regardless, it's low-risk, so self-r+.

PR #244, opened 2017-06-27
Bug 1376501: Add counters to health check endpoints

willkg:


willkg:
This is super cosmetic. self-r+.

PR #246, opened 2017-07-19
Bug 1381713: fix indexerror in is_thunderbird_seamonkey rule

willkg:
This makes sure there's a non-empty ProductName before checking the first
character.

### DIFF
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -245,7 +245,7 @@ mozilla_rules = [
     Rule(
         rule_name='is_thunderbird_seamonkey',
         key='ProductName',
-        condition=lambda x: x[0] in 'TSC',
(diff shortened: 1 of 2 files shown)


antenna/throttler.py
```diff
@@ -245,7 +245,7 @@ def always_match(crash):
     Rule(
         rule_name='is_thunderbird_seamonkey',
         key='ProductName',
-        condition=lambda x: x[0] in 'TSC',
+        condition=lambda x: x and x[0] in 'TSC',
```
willkg: This enforces that x is a truthy value. So None (it shouldn't ever be None at this point anyhow because of the way the throttler works) and empty string are falsey, so this will return False in those situations rather than kicking up an IndexError because there is no first character.

willkg:
Looks good.

PR #247, opened 2017-07-29
Rework docs scripts to build with right permissions

willkg:
This redoes "make docs" so that the resulting files have the permissions of the
host user which makes life soooooo much easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 ANTENNA_ENV ?= "dev.env"
 DC := $(shell which docker-compose)
-UIDGID := $(shell id -u):$(shell id -g)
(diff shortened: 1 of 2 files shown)


willkg:
@adngdb This is what I was thinking of doing. It solves the problem where certain actions create files that have the wrong permissions from the point of view of the host computer. I'd have to either set the permissions to 777 or I'd have to remove files as root. Blech.

I don't think this solves the same problem you're solving with gosu, though.

willkg:
Going to land this--it's soooo much better. The whole permissions conflict thing was so annoying.

adngdb:
:+1:

PR #248, opened 2017-12-23
bug 1426949 - reject crashes from certain browser configurations

willkg:


willkg:
Things to verify:

1. Does the logic for `match_infobar_true` look right?
2. Are the two places in the code that we'll need to change when we do releases easy to identify and is it clear what that person will need to change?

Otherwise, this change is pretty straight-forward. Let me know if it's not.

tests/unittest/test_throttler.py
```diff
@@ -90,6 +189,16 @@ def test_hangid(self):
         throttler = Throttler(ConfigManager.from_dict({}))
         assert throttler.throttle(raw_crash) == (REJECT, 'has_hangid_and_browser', None)
 
+    def test_infobar(self):
+        raw_crash = {
+            'ProductName': 'Firefox',
+            'SubmittedFromInfobar': 'true',
+            'Version': '52.0.2',
```
lonnen: glad to see a more complex version number than major + minor

lonnen:
r+

antenna/throttler.py
```diff
@@ -178,6 +178,31 @@ def always_match(crash):
...
+        return False
+
+    return (
+        product == 'Firefox' and
+        infobar == 'true' and
+        version.startswith(('52.', '53.', '54.', '55.', '56.', '57.', '58.', '59.')) and
+        # FIXME(willkg): buildid is a string that starts with a YYYYMMDD date
+        # so we're doing string compares. Change this to the buildid date when
+        # we push out fixes.
+        buildid < '20180110'
```
lonnen: consider putting this a bit further out. 59 will go live 2018-03-13

lonnen: just in case

tests/unittest/test_throttler.py
willkg: I can sprinkle in a few more.

antenna/throttler.py
willkg: I'll push it further out.

willkg:
I forgot how much I like working on Antenna. CI runs so much faster.

willkg:
I'm going to land this and verify it on -stage.

PR #249, opened 2017-12-24
bug 1426949 - rework throttle code

willkg:
This reworks the throttle code such that if the throttler rejects the crash, that's the end of it--nothing can override a reject.

The diff for this seems involved, but all it does is move the `self.throttler.throttle()` call to the top and if that returns REJECT, that's the end of that. Beyond that, this cleans up some code so that it flows through better.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -288,50 +288,54 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
         :returns tuple: ``(result, rule_name, percentage)``
 
         """
-        # If the raw_crash has a uuid, then that implies throttling, so return
-        # that.
-        if 'uuid' in raw_crash:
-            crash_id = raw_crash['uuid']
-            if crash_id[-7] in (str(ACCEPT), str(DEFER)):
-                result = int(crash_id[-7])
+        # FIXME(willkg): We should move some of these things into throttle
+        # rules instead of having them here.
+
+        # Throttle the crash for an initial result
+        result, rule_name, throttle_rate = self.throttler.throttle(raw_crash)
+
+        # If we didn't reject the crash, then go through it and override the
+        # throttle result in certain circumstances.
+        if result != REJECT:
+            if 'uuid' in raw_crash:
+                # If the raw_crash has a uuid, then that implies throttling, so return
+                # that
+                crash_id = raw_crash['uuid']
+                if crash_id[-7] in (str(ACCEPT), str(DEFER)):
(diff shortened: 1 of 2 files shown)


antenna/breakpad_resource.py
```diff
@@ -288,51 +288,55 @@ def get_throttle_result(self, raw_crash):
...
-            except ValueError:
-                # If we've gotten a ValueError, it means one or both of the
-                # values is bad and we should ignore it and move forward.
-                mymetrics.incr('throttle.bad_throttle_values')
-
-        # If we have a Throttleable=0, then return that.
-        if raw_crash.get('Throttleable', None) == '0':
-            # If the raw crash has ``Throttleable=0``, then we accept the
-            # crash.
-            mymetrics.incr('throttleable_0')
```
willkg: I don't know why this did a separate metrics ping. It gets done again later as a tag which we have in our dashboards.

tests/unittest/test_breakpad_resource.py
```diff
@@ -204,6 +204,12 @@ def test_existing_uuid(self, client):
         # throttleable is not 0
         ({'Throttleable': '1'}, (0, 'has_comments', 100)),
         ({'Throttleable': 'foo'}, (0, 'has_comments', 100)),
+
+        # throttler rejects the code and it stays rejected regardless of throttleable value
+        (
+            {'Throttleable': '0', 'HangID': '555', 'ProcessType': 'browser'},
```
willkg: The hangid rule is easier to simulate and it produces a REJECT, so I used that here. Any REJECT rule would do fine.

antenna/breakpad_resource.py
```diff
@@ -288,51 +288,55 @@ def get_throttle_result(self, raw_crash):
...
+                    # If we've gotten a ValueError, it means one or both of the
+                    # values is bad and we should ignore it and move forward
+                    mymetrics.incr('throttle.bad_throttle_values')
+
+            elif raw_crash.get('Throttleable', None) == '0':
+                # If the raw crash has ``Throttleable=0``, then it was
+                # submitted manually, so accept the crash instead of the
+                # throttle result
+                result = ACCEPT
+                rule_name = 'THROTTLEABLE_0'
```
willkg: This block of code diffs poorly because it's all indented one level under a new `if not REJECT`. Sorry. :(

willkg:
I talked to @lonnen and we're not going to go forward with this. That code could use some tlc since there's effectively two sets of rules and there should only be one. It'd be good to clean that up.

Closing this out.

PR #251, opened 2018-02-05
fixes bug 1433316 - unthrottle Firefox pre-57

willkg:


willkg:
@lonnen Does this rule match what you were thinking?

lonnen:
Yes. This looks correct to me. We can scale it up from there as we are sure that the load isn't an issue.

unrelated to this change, but proximal:  `# 100% of PrductName=Fennec` has a typo

willkg:
Typo fixed!

willkg:
Thank you!

PR #252, opened 2018-02-06
fixes bug 1436176 - update localstack image

willkg:


willkg:
This only affects the local dev environment and if it's wrong, we'll just do another PR to fix it--no biggie. I'm going to self-r+ this.

PR #258, opened 2018-03-12
fix bug 1444939 - nix new relic

willkg:
We never really used New Relic for Antenna, so this removes the bits
that were left over from that.


willkg:
Thank you!

PR #260, opened 2018-03-19
Update docs on triggering a crash in Firefox

willkg:
This updates the tips on how to trigger a crash (and a crash report) in
Firefox.

### DIFF
--- a/docs/breakpad_reporting.rst
+++ b/docs/breakpad_reporting.rst
@@ -155,6 +155,8 @@ development instance. There are a few options:
 
    https://addons.mozilla.org/en-US/firefox/addon/crash-me-now-simple/
 
+   This doesn't work with Firefox 57+.
+
 3. Set environment variables:
 
    https://developer.mozilla.org/en-US/docs/Environment_variables_affecting_crash_reporting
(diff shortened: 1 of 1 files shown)


willkg:
ABRT doesn't work for me anymore, but SIGFPE does. I'm not sure why and didn't spend time looking into it.

Ted's addon doesn't work with Firefox 57+, so I added a note about that.

I also added some text about which processes to kill off which I keep forgetting.

docs/breakpad_reporting.rst
```diff
@@ -163,20 +165,22 @@ development instance. There are a few options:
 
 4. On Linux, you can crash processes using ``kill``::
 
-       kill -ABRT $(pidof firefox)
+       kill -SIGFPE $(pidof firefox)
```
Osmose: `SIGFPE` doesn't work on MacOS for me, but `ABRT` does. You may want to note each for different platforms.

docs/breakpad_reporting.rst
```diff
@@ -163,20 +165,22 @@ development instance. There are a few options:
...
-3. Run ``ps -aef``, find the firefox process id and then do ``kill -ABRT <PID>`` in a
-   third terminal
+3. Run ``ps -aef | grep firefox``, find the process id for the main firefox
+   process or a content process, and then do ``kill -SIGFPE <PID>`` in a third
+   terminal.
 
-4. The Firefox process will crash and the crash report dialog will pop up. Make sure
-   to submit the crash, then click on "Quit Firefox" button
+4. The Firefox process will crash and the crash report dialog will pop up.
+   Make sure to submit the crash, then click on "Quit Firefox" button.
```
Osmose: This process doesn't work for me using my normal Nightly build because crash reporting is disabled over insecure contexts. Can that be disabled?

willkg: I have no idea. I don't think I've had that problem with release or nightly. I think we should push off figuring this out to a separate PR/bug.

willkg:
^^^ Do those edits for the section on killing Firefox work better?

willkg:
@Osmose ^^^ Is this ok?

Osmose:
Whoops, sorry that I missed this. Adding to my todo list now.

willkg:
Thank you!

PR #261, opened 2018-04-16
fix bug 1409427 - switch to sha256 and add MinidumpSha256 checksum

willkg:
This switches `dump_checksum` from md5 hexdigest to sha256 hexdigest. We don't use this currently, but sha256 is better and maybe we'll use it for something someday.

This adds a new MinidumpSha256 field with the sha256 hexdigest of the `upload_file_minidump` file.

This also moves the checksum-figuring code out of `extract_payload` and into the `on_post` handler near other code that adds fields to the raw crash.


willkg:
Thank you!

PR #262, opened 2018-05-18
convert to circleci 2.0

milescrabill:
some caveats:

- this doesn't really use the workflow feature, which is cool but kind of a pain
- we're now also pushing `mozilla/socorro_collector` images because it's convenient for ops


.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
+version: 2
+jobs:
+  build_test_push:
+    docker:
+      - image: docker:stable
+    environment:
+      BUILD_IMAGE: antenna:build
+      TAGGED_IMAGE: mozilla/antenna
```
willkg: Socorro uses this image name, but we can switch it. Is there anything else using this image name?

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+                export TAGGED_IMAGE="${TAGGED_IMAGE}:${CIRCLE_TAG}"
+                export TAGGED_IMAGE_2="${TAGGED_IMAGE_2}:${CIRCLE_TAG}"
+            fi
+
+            if [ "${CIRCLE_BRANCH}" == "master" ]; then
+                echo "${DOCKER_PASSWORD}" | docker login -u="${DOCKER_USERNAME}" --password-stdin
+                docker tag "${BUILD_IMAGE}" "${TAGGED_IMAGE}"
+                docker tag "${BUILD_IMAGE}" "${TAGGED_IMAGE_2}"
+                docker push "${TAGGED_IMAGE}"
+                docker push "${TAGGED_IMAGE_2}"
```
willkg: The `deploy-dockerhub.sh` script would retry pushing, but this doesn't. If this fails, what happens?

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+  # we use filters to whitelist jobs that should be run for tags
+
+  # workflow jobs are run in _all_ branch builds by default
+  # we use filters to blacklist jobs that shouldn't be run for a branch
+
+  # see: https://circleci.com/docs/2.0/workflows/#git-tag-job-execution
+
+  build_test_push:
+    jobs:
+      - build_test_push:
```
willkg: The Socorro `.circleci/config.yml` has `build` here. I'm pretty sure this is just a job name, so it's probably not a big deal, but I mention it in case you want consistency between all the socorro-y things.

.circleci/config.yml
milescrabill: I doubt it, I imagine we're the only consumers of this image.

.circleci/config.yml
milescrabill: If the push fails, CI will fail. This is an unfortunate side-effect. If that happens, Circle will send a failure email, and the build can be restarted.

If we were using workflows to handle this, we could restart that step independently. That creates quite a bit of complexity in the config though, which is why I avoided it. I got rid of `deploy-dockerhub.sh` for the same reason - I can add retry functionality if we see failures.

.circleci/config.yml
milescrabill: Yes, this is the workflow / job name.

PR #266, opened 2018-07-18
fix bug 1447032: update requirements, drop dev-base image, minor fixes

willkg:
This does a few things:

1. Drops the dev-base docker image so we're only using one image now. Figured I'd simplify that out.
2. Splits requirements into defaults and constraints just like we do with Socorro. This simplifies dependency maintenance over time and reduces the likelihood we have libs installed that we're not using.
3. Removed some libs we weren't using.
4. Fixes a minor bug I bumped into in mini poster.
5. Updates requirements and fixes a single test that failed.

This shouldn't require any changes in infrastructure or how we do anything. This should be pretty safe.


willkg:
Best way to test this out is the following:

1. `make build`
2. `make test`
3. `make systemtest`
4. `make docs`

If all that works out, then we're probably fine. We have an integration test that runs post-deploy, so we'll know for certain after it hits stage.

AlexisDeschamps:
I get `copying static files... WARNING: html_static_path entry '/app/docs/_static' does not exist` when doing `make docs`. Not if that is expected. Other stuff seems fine though.

willkg:
Thank you!

PR #267, opened 2018-07-19
fix bug 1476840: re-add datadog library to dependencies

willkg:
I messed up and inadvertently removed it in the dependencies update pass
I did yesterday. This re-adds it. Without it, markus can't send any
metrics to Datadog.


willkg:
Thank you!

PR #268, opened 2018-08-04
fix bug 1480936: change throttle default to REJECT

willkg:
This changes the throttle default from `DEFER` to `REJECT`. So instead of
Socorro saving everything even if it's not going to process it, it will
now only save crash data for crash reports it plans to process.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -19,8 +19,6 @@ import markus
 
 from antenna.heartbeat import register_for_life, register_for_heartbeat
 from antenna.throttler import (
-    ACCEPT,
-    # DEFER,
     REJECT,
     RESULT_TO_TEXT,
     Throttler,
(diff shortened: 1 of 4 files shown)


willkg:
I want to wait on landing this until I get a +1 from some people.

antenna/breakpad_resource.py
```diff
@@ -267,70 +265,16 @@ def extract_payload(self, req):
...
-            rule_name = 'THROTTLEABLE_0'
-            throttle_rate = 100
-
-        else:
-            # At this stage, nothing has given us a throttle answer, so we
-            # throttle the crash.
-            result, rule_name, throttle_rate = self.throttler.throttle(raw_crash)
+        # At this stage, nothing has given us a throttle answer, so we
+        # throttle the crash.
+        result, rule_name, throttle_rate = self.throttler.throttle(raw_crash)
```
willkg: I radically simplified this method by moving things that should have been rules to the rule set in the throttler. That simplifies the mental model for "throttling" since now all the rule-ish things are in one linear rule set.

Yay for progress!

antenna/breakpad_resource.py
```diff
@@ -267,70 +265,16 @@ def extract_payload(self, req):
...
-
-        #         throttle_rate = int(raw_crash['throttle_rate'])
-        #         if not (0 <= throttle_rate <= 100):
-        #             raise ValueError('Throttle rate is not a valid value: %r', result)
-        #         return result, 'ALREADY_THROTTLED', throttle_rate
-
-        #     except ValueError:
-        #         # If we've gotten a ValueError, it means one or both of the
-        #         # values is bad and we should ignore it and move forward.
-        #         mymetrics.incr('throttle.bad_throttle_values')
```
willkg: We had commented this out when we implemented "reject crash reports from unsupported products" so we could test the code on stage. This code is here solely so that stage processes all the incoming crash reports sent by prod. This makes it easier to see some things on stage and debug issues with the processor and webapp.

However, we have good tools for taking crash data from prod and submitting it to stage--we don't need a stage-specific exception and this code is funky and creates other issues and prevents us from testing throttling on stage.

So given all that, I figure it's easier maintenance-wise to just ditch it.

antenna/breakpad_resource.py
```diff
@@ -267,70 +265,16 @@ def extract_payload(self, req):
...
-
-        # If we have a Throttleable=0, then return that.
-        if raw_crash.get('Throttleable', None) == '0':
-            # If the raw crash has ``Throttleable=0``, then we accept the
-            # crash.
-            mymetrics.incr('throttleable_0')
-            result = ACCEPT
-            rule_name = 'THROTTLEABLE_0'
-            throttle_rate = 100
```
willkg: This got moved to a throttler rule.

antenna/throttler.py
```diff
@@ -113,11 +113,11 @@ def throttle(self, raw_crash):
                 if rule.percentage == 100 or (random.random() * 100.0) <= rule.percentage:  # nosec
                     response = ACCEPT
                 else:
-                    response = DEFER
+                    response = REJECT
```
willkg: This is the big change. Instead of "ACCEPT if we roll a good number and DEFER if we don't", we now "ACCEPT if we roll a good number and REJECT if we don't".

antenna/throttler.py
```diff
@@ -113,11 +113,11 @@ def throttle(self, raw_crash):
...
                 return response, rule.rule_name, rule.percentage
 
         # None of the rules matched, so we defer
-        return DEFER, 'NO_MATCH', 0
+        return REJECT, 'NO_MATCH', 0
```
willkg: This is also a change--if the crash report matched none of the rules, then we reject it.

However, it has no effect on anything since we added a catch-all "accept everything" rule to the end.

Why do it that way? I wanted to make it easier to think about and for other groups to use Antenna. I want to get rid of the notion of DEFER altogether. Seems like if a crash report matches none of the rules set by some non-Mozilla group, it makes more sense to reject it.

antenna/throttler.py
```diff
@@ -277,86 +277,79 @@ def match_unsupported_product(throttler, data):
...
     ),
 
-    # 100% of crashes that have a comment
+    # Accept 100% of crash reports submitted through about:crashes
+    Rule(
+        rule_name='throttleable_0',
+        key='Throttleable',
+        condition=lambda throttler, x: x == '0',
+        percentage=100
+    ),
```
willkg: This is the rule-ified Throttleable rule.

antenna/throttler.py
```diff
@@ -277,86 +277,79 @@ def match_unsupported_product(throttler, data):
...
     ),
 
-    # 100% of all ReleaseChannel=nightly
+    # Accept 100% of crash reports in ReleaseChannel=nightly
     Rule(
         rule_name='is_nightly',
         key='ReleaseChannel',
         condition=lambda throttler, x: x.startswith('nightly'),
         percentage=100
     ),
```
willkg: I kept the above two rules because I (ab)use the dashboard showing throttle rules that got triggered to figure out what happened when we get spikes of incoming crashes. I shouldn't do that. I should use different data. Can't really use super search because I want to view a timeseries--not counts. I'll think about this more at some point. For now, I'm going to leave these two rules in.

antenna/throttler.py
```diff
@@ -277,86 +277,79 @@ def match_unsupported_product(throttler, data):
...
-        key='ProductName',
-        condition=lambda throttler, x: x and x[0] in 'TSC',
+        rule_name='accept_everything',
+        key='*',
+        condition=always_match,
         percentage=100
-    ),
-
-    # Reject everything else
+    )
```
willkg: I added this "accept everything" catch all since it means that all "how does the Mozilla Socorro collector throttle?" questions can be answered solely by looking at this rule set.

willkg:
I got the +1 to go ahead. I re-wrote the code changes. I reviewed them and they look ok.

I'm going to let this sit a bit and then look through it again with fresher eyes.

antenna/throttler.py
willkg: This would be better codified as part of the rule structure. I'll look at doing that now.

willkg:
I went through this again and it looks good. I'm going to land it and test it further on stage. self-r+

PR #269, opened 2018-08-09
fix bug 1476941: reject unsupported products

willkg:
This switches the throttler to reject unsupported products. We'll add
products to this list over time as requested.

This also cleans up the throttler. It gets rid of the weirdness around
conditions, so now all conditions are functions. It gets rid of the
one regexp--we could do the same thing without a regexp. It adds
documentation.

### DIFF
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -30,8 +30,6 @@ RESULT_TO_TEXT = {
     2: 'REJECT'
 }
 
-REGEXP_TYPE = type(re.compile(''))
-
 
 def parse_attribute(val):
     module, attribute_name = val.rsplit('.', 1)
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -268,11 +268,11 @@ MOZILLA_RULES = [
     # supported products is empty
     Rule(
         rule_name='unsupported_product',
-        key='ProductName',
+        key='*',
         condition=(
-            lambda throttler, val: (
+            lambda throttler, data: (
                 throttler.config('products') and
-                val not in throttler.config('products')
+                data.get('ProductName', '') not in throttler.config('products')
             )
         ),
         percentage=None
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -2,6 +2,13 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+"""Defines the throttler for the breakpad handler for Antenna.
+
+The throttler lets you define rules for which crashes to accept (save and
+process), defer (save, but not process), and reject.
+
+"""
+
 import importlib
 import logging
 import random
@@ -47,8 +54,11 @@ class Throttler(RequiredConfigMixin):
 
     If you don't want to throttle anything, use this::
 
-        THROTTLE_RULES=antenna.throttler.accept_all
+        THROTTLE_RULES=antenna.throttler.ACCEPT_ALL
 
+    If you want to support all products, use this::
+
+        PRODUCTS=antenna.throttler.ALL_PRODUCTS
 
     To set up a rule set, put it in a Python file and define the rule set
     there. For example, you could have file ``myruleset.py`` with this in it::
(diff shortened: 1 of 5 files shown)


antenna/throttler.py
```diff
@@ -72,10 +82,16 @@ class Throttler(RequiredConfigMixin):
...
+        default='antenna.throttler.MOZILLA_RULES',
         doc='Python dotted path to ruleset',
         parser=parse_attribute
     )
+    required_config.add_option(
+        'products',
+        default='antenna.throttler.MOZILLA_PRODUCTS',
+        doc='Python dotted path to list of supported products',
+        parser=parse_attribute
+    )
```
willkg: I made this configurable because I secretly hope there are other people using Antenna out there.

antenna/throttler.py
```diff
@@ -90,7 +106,7 @@ def throttle(self, raw_crash):
 
         """
         for rule in self.rule_set:
-            match = rule.match(raw_crash)
+            match = rule.match(self, raw_crash)
```
willkg: In order for rule conditions to use configuration, I needed to pass the configuration in as an argument. Instead of doing that, I passed the throttler. I figure this lets conditions do other things, too.

Further, I decided this reads poorly:

```
lambda config, x: ...
```

Config? What config? There's lots of config everywhere.

Instead, it's this:

```
lambda throttler, x: throttler.config('something')
```

That feels a lot clearer to me.

antenna/throttler.py
```diff
@@ -138,7 +154,9 @@ def __init__(self, rule_name, key, condition, percentage):
         """
         self.rule_name = rule_name
         self.key = key
-        self.condition = self._to_handler(condition)
+        if not callable(condition):
+            raise ValueError('condition %r is not callable' % condition)
+        self.condition = condition
```
willkg: I decided I hated how the condition stuff read. I could never remember how it got converted and it was so special-casey. So instead of that, all conditions are functions. That's easy to read and remember and lets us get rid of some code I was never particularly fond of. Most conditions were functions already anyhow.

antenna/throttler.py
```diff
@@ -138,7 +154,9 @@ def __init__(self, rule_name, key, condition, percentage):
         """
         self.rule_name = rule_name
         self.key = key
-        self.condition = self._to_handler(condition)
+        if not callable(condition):
+            raise ValueError('condition %r is not callable' % condition)
```
willkg: This will never happen in production. However, it makes it muuuuuch easier to debug code and it makes it more likely all our issues will pop up when running the tests.

antenna/throttler.py
```diff
@@ -223,29 +225,61 @@ def match_firefox_pre_57(data):
...
+        rule_name='unsupported_product',
+        key='ProductName',
+        condition=(
+            lambda throttler, val: (
+                throttler.config('products') and
+                val not in throttler.config('products')
+            )
+        ),
+        percentage=None
+    ),
```
willkg: Some day, I'm going to fix the whole "`percentage=None` means reject this" nonsense. That makes no sense. But that day is not today.

So what this does is check if supported products are configured and if so, if the `ProductName` value is in there. If not, then this rule will reject the crash report.

antenna/throttler.py
```diff
@@ -223,29 +223,61 @@ def match_firefox_pre_57(data):
...
+ALL_PRODUCTS = []
+
+
+#: List of supported products; these have to match the ProductName of the
+#: incoming crash report
+MOZILLA_PRODUCTS = [
+    'Firefox',
+    'Fennec',
+    'Thunderbird',
+    'SeaMonkey'
```
willkg: This is the list of products we currently support. "What about FennecAndroid?" you might say. To that, I reply, "Ugh. FennecAndroid is an abomination of MacGuyvering the system to bend it to the will of eldritch darkness." Then I might sigh, pause a bit, and recount my last expedition through the arctic waters of the processor rules which includes the run-in with the ProductRewrite rule which converts incoming crash reports with ProductName of Fennec to FennecAndroid. "It's an illusion! Clown shoes all the way down! Parrots! Too many parrots!" I scream! Roll 2d10 damage from psychic horror.

antenna/throttler.py
```diff
@@ -223,29 +223,61 @@ def match_firefox_pre_57(data):
...
         key='*',
         condition=match_infobar_true,
         percentage=None
     ),
 
+    # Reject all unsupported products; this does nothing if the list of
+    # supported products is empty
+    Rule(
+        rule_name='unsupported_product',
+        key='*',
```
willkg: Have to use `'*'` here because otherwise the match code will skip this rule if there's no `ProductName`. We want it to run this rule and if there's no `ProductName`, reject the crash report.

willkg:
I think this looks good, but this is a high risk change, so I'm going to sit on it and re-review it tomorrow.

willkg:
Still looks good. I wrote up a rough test plan for stage to verify things are ok. self-r+

PR #270, opened 2018-08-10
bug 1476941: nix UUID check

willkg:


tests/unittest/test_s3_crashstorage.py
```diff
@@ -41,7 +41,7 @@ def test_crash_storage(self, client, s3mock):
         )
         data, headers = multipart_encode({
             'uuid': 'de1bb258-cbbf-4589-a673-34f800160918',
-            'ProductName': 'Test',
+            'ProductName': 'Thunderbird',
```
willkg: These tests would have broken with the throttling changes, except that they include a uuid which has an accept/defer character which skips throttling.

willkg:
self-r+

PR #272, opened 2018-08-13
Add docs requirements files

willkg:
ReadTheDocs can only handle one requirements file (as far as I can see), so
this creates one that points at the two files we have in this repo.


willkg:
This only affects RTD.

PR #275, opened 2018-08-21
bug 1483719: add support for GeckoView

willkg:


willkg:
This is a trivial change. Will test it on stage to verify. self-r+

PR #276, opened 2018-08-23
fix bug 1440031: remove pre-57 throttle rule

willkg:
We had a rule for throttling pre-57 at 20% rather than the normal 10%.
This helped us catch up with our data after deleting everything.

We no longer need this, so it's time to remove it. This commit does that.


willkg:
This is pretty straight-forward removal of this rule. self-r+

PR #278, opened 2018-10-18
fix bug 1500243: add fakeaccept for b2g

willkg:
B2G products send crash reports and will retry ad infinitum on any non-accept.
This fixes the throttler to handle this explicitly by "fake accepting" the
crash. This involves sending an accept-like response, but throwing the crash
away.

We want to be super careful doing this because we're throwing crash data away.
It's ok in this case because we don't want B2G crash data and we don't want
B2G devices to be trying to resend over and over again.


willkg:
This looks good. It's not very invasive, it solves the immediate problem, and if we have a better solution later, we can go with that. This doesn't put us in a worse position. This is easily tested. self-r+

PR #279, opened 2018-10-29
bug 1501298: add ReferenceBrowser to supported products

willkg:


willkg:
I verified the spelling is correct. self-r+

PR #288, opened 2019-01-31
Update Makefile to be like Socorro's

willkg:
This moves `.PHONY` markings to just before the relevant rule. This reduces
accidental missing `.PHONY` markings.

This also redoes default and help so that the default is to be helpful
rather than tell you to ask for help.


willkg:
self-r+

PR #290, opened 2019-02-12
fix bug 1527348: add metrics for de_null

willkg:


willkg:
FYI: This adds a lot of logging in the local dev environment, so I think we're going to want to either:

1. make this temporary, or
2. change it to a gauge so it creates a "same" and "different" once per crash

Mmm... actually, I like that better--going to change it now.

willkg:
self-r+

PR #291, opened 2019-02-12
fix bug 1357246: remove de-nulling code

willkg:
This removes the de-nulling code from Antenna. It'll now be done in the
processor.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -26,7 +26,6 @@ from antenna.throttler import (
 )
 from antenna.util import (
     create_crash_id,
-    de_null,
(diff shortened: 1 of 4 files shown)


willkg:
I just pushed code to capture metrics on de-nulling. I want to wait until that's run for a day before doing this.

willkg:
I added metrics for de-nulling and they've been out for a little less than a day. In all that time, de-nulling hasn't had any effect.

So, let's move forward with this and remove de-nulling from Antenna. If there are problems, we have the crash report so we can fix the processor and then reprocess the crash.

willkg:
This looks good. self-r+

PR #292, opened 2019-02-14
fix bug 1527919: use console logging in a local dev environment

willkg:
This fixes logging so it uses mozlog in a server environment and console
in a local dev environment. This makes logs muuuuuch easier to read
when working on things.


willkg:
I'll check stage after this deploys to make sure everything worked out there, too. self-r+

PR #293, opened 2019-02-19
fix bug 1529026: make verification a separate step

willkg:
Parts of Antenna verify that they're configured correctly at instantiation time. This changes that so that everything gets instantiated and then everything that needs to verify something gets a chance to verify. This makes Antenna startup issues easier to debug.

### DIFF
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -237,7 +237,7 @@ class AntennaAPI(falcon.API):
         self.hb = HeartbeatManager(config)
 
     def add_route(self, name, uri_template, resource, *args, **kwargs):
-        """Adds specified Falcon route
+        """Add specified Falcon route.
 
         :arg str name: friendly name for this route; use alphanumeric characters
 
@@ -250,7 +250,7 @@ class AntennaAPI(falcon.API):
         super().add_route(uri_template, resource, *args, **kwargs)
 
     def get_resource_by_name(self, name):
-        """Returns the registered resource with specified name
+        """Return registered resource with specified name.
 
         :arg str name: the name of the resource to get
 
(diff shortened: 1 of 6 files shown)


antenna/app.py
```diff
@@ -49,6 +49,9 @@ def filter(self, record):
             }
         },
         'formatters': {
+            'socorroapp': {
+                'format': '%(asctime)s %(levelname)s - %(name)s - %(message)s',
+            },
```
willkg: This fixes logging formatting so when you're running it in the local dev environment, you can see which logger logged what.

antenna/app.py
```diff
@@ -252,24 +260,31 @@ def get_resource_by_name(self, name):
...
     def get_runtime_config(self, namespace=None):
+        """Return generator of runtime configuration for all resources."""
         for res in self.get_resources():
             if hasattr(res, 'get_runtime_config'):
                 for item in res.get_runtime_config(namespace):
                     yield item
 
+    def verify(self):
+        """Verify that Antenna is ready to start."""
+        self.hb.verify()
```
willkg: The heartbeat system already had registration for hooks, so I decided to make verification part of the heartbeat stuff.

antenna/app.py
```diff
@@ -302,17 +317,22 @@ def get_app(config=None):
...
     app.add_route('version', '/__version__', VersionResource(config, basedir=app_config('basedir')))
     app.add_route('heartbeat', '/__heartbeat__', HeartbeatResource(config, app))
     app.add_route('lbheartbeat', '/__lbheartbeat__', LBHeartbeatResource(config))
     app.add_route('broken', '/__broken__', BrokenResource(config))
 
+    # Finish logging configuration
     log_config(logger, app)
 
+    # Verify that we're ready to start
+    app.verify()
```
willkg: Now all verification happens after everything has been created but before Antenna is handling incoming connections.

Exceptions that are kicked up here will prevent Antenna from starting and will get sent to Sentry since Sentry is configured to listen to logged exceptions.

willkg:
Seems fine. I'll check it out on stage. self-r+

PR #294, opened 2019-02-20
Fix pydocstyle lint issues

willkg:
This fixes issues raised by flake8-docstrings which uses pydocstyle. I
didn't add it to the "make lint" pass because I'm on the fence about whether
to enforce it for everyone. However, it did make sense to run it once
and fix any issues and thus improve the docs.


willkg:
These are all docstring changes. The results should make docstrings which are extracted into the docs on ReadTheDocs more consistent. self-r+

PR #295, opened 2019-03-04
fix bug 1531865: redo tests and configuration

willkg:
This redoes how tests are run so that they can be run against services rather than require services to be mocked out.

This also redoes how configuration is done so that it matches what we do in Socorro. I think while there's more scaffolding required, it's more straight-forward for maintainers and their typical use cases to use.


antenna/app.py
```diff
@@ -292,8 +291,6 @@ def get_app(config=None):
     if config is None:
         config = ConfigManager(
             environments=[
-                # Pull configuration from env file specified as ANTENNA_ENV
-                ConfigEnvFileEnv([os.environ.get('ANTENNA_ENV')]),
                 # Pull configuration from environment variables
                 ConfigOSEnv()
```
willkg: I could probably simplify this further and just do `ConfigManager.basic_config()`. That sets up a `ConfigManager` with `ConfigEnvFileEnv('.env')` and `ConfigOSEnv()`. We pull in `my.env` via docker-compose and I didn't think I wanted to add the `.env` thing to the server environments--there should be only one way to configure things.

Anyhow, so I didn't do that and set it up explicitly like this.

docker/Dockerfile
```diff
@@ -1,7 +1,12 @@
-FROM python:3.6.2-slim
+FROM python:3.6.8-slim
+
+# Set up user and group
+ARG groupid=10001
+ARG userid=10001
```
willkg: This makes it easier to build docs when running Docker on Linux.

willkg:
This works locally. I read through the CircleCI output and I'm pretty sure it worked there, too. I'm going to land this and make sure it deploys to stage correctly. If that works, then I think we're good here. self-r+

PR #296, opened 2019-03-04
bug 1517807: Fix s3 test class name; remove pytest-env

willkg:
The s3 test class names were weird and there were two classes but only need to be one. This fixes both of those issues.

Antenna isn't using pytest-env, so best to remove it.

This doesn't redo the s3mock stuff altogether, but does fix some minor things with it.


willkg:
This is minor. I pulled it out of the pubsub changes since it's not related to those changes. self-r+

PR #297, opened 2019-03-04
bug 1531865: fix image name in build/push step

willkg:


willkg:
This fix affects the build/push, so the way to test it is by landing this. self-r+.

PR #299, opened 2019-03-06
fix bug 1527343: implement publishing to Pub/Sub

willkg:
This implements publishing to Pub/Sub. It's a minor rewrite of parts of Antenna, so it's worth really going through it.

This defaults to using the `NoOpCrashPublish` class which will just log what it's "publishing". We want to use that for now until I've implemented the processor and webapp sides. When we've done that, then we'll work out the migration for stage, test it fully there, and then do the same for prod.


Makefile
```diff
@@ -63,7 +63,7 @@ run: my.env .docker-build
 
 .PHONY: shell
 shell: my.env .docker-build
-	${DC} run base bash
+	${DC} run web bash
```
willkg: This should run in the `web` container since that has the environment variables we need.

antenna/app.py
```diff
@@ -137,6 +137,21 @@ def log_config(logger, component):
...
+            # Pull configuration from environment variables
+            ConfigOSEnv()
+        ],
+        doc=(
+            'For configuration help, see '
+            'https://antenna.readthedocs.io/en/latest/configuration.html'
+        )
+    )
+
+    return config
```
willkg: I extracted this, but forget why. I suspect there was a reason for it, but then the reason went away. Maybe it got a job and moved to a big city somewhere. Maybe it's decided to be an astronaut. Space. Spaaaaaaaace.

antenna/breakpad_resource.py
```diff
@@ -124,6 +136,9 @@ def __init__(self, config):
         # Queue for crashmover of crashes to save
         self.crashmover_save_queue = deque()
 
+        # Queue for publishing crashids to processor
+        self.crashmover_publish_queue = deque()
```
willkg: This is effectively a two-step crashmoving process now. I thought about generalizing that to an n-step process but decided to wait on that until it's 3+-steps.

antenna/breakpad_resource.py
```diff
@@ -368,45 +393,76 @@ def on_post(self, req, resp):
...
+                    )
+
+                    # After MAX_ATTEMPTS, we give up on this crash and move on
+                    if crash_report.errors < MAX_ATTEMPTS:
+                        self.crashmover_publish_queue.append(crash_report)
+                    else:
+                        logger.error(
+                            '%s: too many errors trying to publish; dropped',
+                            crash_report.crash_id
+                        )
```
willkg: If we ever get to 3+-steps, we can generalize this.

antenna/ext/crashpublish_base.py
```diff
@@ -0,0 +1,57 @@
...
+
+    required_config = ConfigOptions()
+
+    def __init__(self, config):
+        self.config = config.with_options(self)
+
+    def publish_crash(self, crash_id):
+        """Publish the crash id
+
+        FIXME(willkg): How should this method handle exceptions?
```
willkg: FIXME: Deal with this FIXME.

antenna/ext/crashpublish_base.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        raise NotImplementedError
+
+
+class NoOpCrashPublish(CrashPublishBase):
+    """This is a no-op crash publish class that logs crashes it would have published.
+
+    It keeps track of the last 10 crash ids in ``.crashes`` instance attribute
+    with the most recently published crash id at the end of the list. This
+    helps when writing unit tests for Antenna.
```
willkg: I didn't end up using this at all in the tests. I'll look into that. I either need to add some more test stuff, or take this out.

antenna/ext/pubsub/crashpublish.py
```diff
@@ -0,0 +1,75 @@
...
+        super().__init__(config)
+
+        self.project_id = self.config('project_id')
+        self.topic_name = self.config('topic_name')
+
+        # Batches crash_ids for up 1024 bytes or a second
+        self.batch_settings = pubsub_v1.types.BatchSettings(
+            max_bytes=1024,
+            max_latency=1,
+        )
```
willkg: Since crash ids are small, I think we won't hit `max_bytes`. I think we'll hit the 1-second latency for batching. That's probably fine. If we find we want to tune that, we can make it configurable.

antenna/ext/pubsub/crashpublish.py
```diff
@@ -0,0 +1,75 @@
...
+        self.topic_path = self.publisher.topic_path(self.project_id, self.topic_name)
+
+        register_for_verification(self.verify_topic)
+
+    def verify_topic(self):
+        """Verify topic exists and can be viewed.
+
+        NOTE(willkg): This requires View permissions and doesn't actually
+        guarantee we can write to the topic. To do that, we'd change what's in
+        the queue and that's not great.
```
willkg: I need to lift the permission requirements to the class docstring so it's in the documentation.

antenna/ext/pubsub/crashpublish.py
```diff
@@ -0,0 +1,75 @@
...
+
+        """
+        self.publisher.get_topic(self.topic_path)
+
+    # FIXME(willkg): In what circumstances should this retry? Also, how
+    # does failure work with batching?
```
willkg: I think the Pub/Sub library is doing some retry work. It's happening in another thread, though, so we never find out if it fails.

@sciurus and I were talking about something similar and decided it's worth writing a script that compares raw crashes to processed crashes in the S3 bucket for a given day and toss anything missing in the processing queue. If we had that (which we should have for a variety of reasons), then we can probably be more lax in the publishing side of things.

antenna/gunicornhooks.py
```diff
@@ -2,6 +2,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+# Monkey patch stdlib with gevent-friendly bits--need to do this here
+# before we import anything else. Gunicorn does this too late.
+from gevent import monkey
+
+monkey.patch_all()
```
willkg: The Pub/Sub library uses threads to batch and send asynchronously. We need to gevent patch that before it gets loaded so it's coroutines and not threads. The gunicorn bootstrapping cycle does that too late, hence adding this here.

bin/pubsub_cli.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python
```
willkg: This is literally the worst cli I've written. It's mediocre at best. But it works and I didn't want to spend more time on it at this point. I figure we can hone it as needs arise.

docker-compose.yml
```diff
@@ -51,6 +52,17 @@ services:
     links:
       - web
 
+  # Google Pub/Sub emulator
+  pubsub:
+    build:
+      context: pubsub
+      dockerfile: Dockerfile
+    environment:
+      - PUBSUB_PROJECT_ID=socorro
```
willkg: Need to add a note that this needs to match `docker/config/local_dev.env`. Also, I think it might not match now--should check that.

pubsub/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+# Install Java and a pubsub-emulator
+RUN apk --update add openjdk8-jre
+RUN gcloud components install pubsub-emulator beta --quiet
+
+# Copy the run script
+COPY run_pubsub.sh /tmp
+
+EXPOSE 5010
+
+CMD /tmp/run_pubsub.sh
```
willkg: We should move this to `docker/Dockerfile.pubsub`.

pubsub/run_pubsub.sh
```diff
@@ -0,0 +1,14 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Runs pubsub emulator inside the pubsub container
+
+# Set up project and run pubsub
+mkdir -p /tmp/data
+gcloud config set project "${PUBSUB_PROJECT_ID}"
+gcloud beta emulators pubsub start \
+    --data-dir=/tmp/data \
+    --host-port=0.0.0.0:${PUBSUB_PORT}
```
willkg: We should move this to `docker/`.

tests/systemtest/test_post_crash.py
```diff
@@ -60,6 +55,15 @@ def verify_crash_data(self, crash_id, raw_crash, dumps, s3conn):
...
             assert key in s3conn.list_objects(prefix=key)
 
+    def verify_published_data(self, crash_id, pubsub):
+        # Verify crash id was published--this might pick up a bunch of stuff,
+        # so we just verify it's one of the things we picked up
+        if 'PUBSUB_EMULATOR_HOST' in os.environ:
+            crash_ids = [crash_id.decode('utf-8') for crash_id in pubsub.list_crashids()]
+            assert crash_id in crash_ids
+        else:
+            print('SKIPPING PUBLISH CHECK--NOT USING EMULATOR')
```
willkg: My problem with this line is that this is not seen. So the Pub/Sub stuff will not run, but does it silently. It'd be nice if it didn't run, but in a more visible manner somehow.

antenna/breakpad_resource.py
```diff
@@ -368,45 +393,76 @@ def on_post(self, req, resp):
             resp.body = 'CrashID=%s%s\n' % (self.config('dump_id_prefix'), crash_id)
 
     def hb_run_crashmover(self):
-        """Check to see if it should spawn a crashmover and does if appropriate."""
+        """Spawn a crashmover if there's work to do."""
         # Spawn a new crashmover if there's stuff in the queue and there isn't
         # one currently running
         if self.crashmover_save_queue and self.crashmover_pool.free_count() > 0:
```
willkg: Oops--this needs to factor in the `self.crashmover_publish_queue`, too.

Mmm... part of me wants to generalize the queue thing now.

antenna/breakpad_resource.py
```diff
@@ -368,45 +393,76 @@ def on_post(self, req, resp):
...
-                    mymetrics.incr('save_crash_dropped.count')
+        while self.crashmover_save_queue or self.crashmover_publish_queue:
+            if self.crashmover_save_queue:
+                crash_report = self.crashmover_save_queue.popleft()
+
+                try:
+                    # Save crash and then toss crash_id in the publish queue
+                    self.crashmover_save(crash_report)
+
+                    crash_report.errors = 0
```
willkg: Once we've successfully saved the crash report, we zero out the errors and move it to publishing.

bin/pubsub_cli.py
jgmize: I think you're being too hard on yourself here. It's at least [premium mediocre](https://www.ribbonfarm.com/2017/08/17/the-premium-mediocre-life-of-maya-millennial/).

willkg: Ha!

willkg: Also, why are you reading the self-review pull request comments from a deranged developer on a solo project doing yet another Sisyphusian infrastructure migration? I am concerned for your health.

jgmize: The concern is mutual, my friend. Don’t hesitate to reach out if you’d like to chat.

antenna/gunicornhooks.py
willkg: This need is moot now, but I think I'm going to keep this here.

docker/Dockerfile.pubsub
```diff
@@ -0,0 +1,12 @@
+FROM google/cloud-sdk:234.0.0-alpine
```
willkg: They're up to 237 now. Might as well update.

docker/config/local_dev.env
```diff
@@ -15,10 +15,19 @@ STATSD_NAMESPACE=mcboatface
...
 CRASHSTORAGE_ENDPOINT_URL=http://localstack-s3:5000
 CRASHSTORAGE_REGION=us-east-1
 CRASHSTORAGE_ACCESS_KEY=foo
 CRASHSTORAGE_SECRET_ACCESS_KEY=foo
 CRASHSTORAGE_BUCKET_NAME=antennabucket
+
+# Pub/Sub things for local development
+CRASHPUBLISH_PROJECT_ID=local_dev_socorro
+CRASHPUBLISH_TOPIC_NAME=local_dev_socorro_normal
+CRASHPUBLISH_SUBSCRIPTION_NAME=local_dev_socorro_sub
```
willkg: The third one here is for the pubsub-cli. It's not otherwise used at all.

willkg:
This seems good. The plan for this PR is this:

1. land it, but use the default configuration which uses the `NoOpCrashPublish` class
2. watch it on stage

That tests out a bunch of the changes and we can let it sit over the weekend.

We'll switch it to publish to Pub/Sub when I finish the changes required in Socorro.

self-r+

PR #300, opened 2019-03-07
fix bug 1359147: fix s3 and pubsub startup and health checks

willkg:
This fixes the `S3CrashStorage` startup check to verify that it can write to the S3 bucket instead of just HEADing the bucket which is a different permission.

This also adds a `PubSubCrashPublish` health check which verifies it can access the topic.

### DIFF
--- a/antenna/ext/s3/connection.py
+++ b/antenna/ext/s3/connection.py
@@ -6,6 +6,7 @@
 import io
 import logging
 import random
+import uuid
 
 import boto3
 from botocore.client import ClientError, Config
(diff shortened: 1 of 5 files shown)


willkg:
Once this lands, we'll want to add an expiration to the S3 buckets `test/` path so that it doesn't accumulate test files.

antenna/ext/s3/connection.py
```diff
@@ -172,24 +172,23 @@ def _build_client(self):
...
-        either need to orphan a gazillion files or we'd also need DELETE
-        permission.
+        This will do multiple attempts and then give up and throw an exception.
 
         """
-        self.client.head_bucket(Bucket=self.bucket)
-
-    def _create_bucket(self):
-        # NOTE(willkg): Don't use this except with a fake s3 and dev environments.
-        self.client.create_bucket(Bucket=self.bucket)
```
willkg: I took this out since nothing should be calling this.

willkg:
self-r+

PR #301, opened 2019-03-12
fix bug 1533732: add support for JSON-encoded fields

willkg:
This adds support for encoding the raw crash extra bits as a single
JSON-encoded value rather than as a series of key-value pairs in
the multipart/form-data serialized HTTP payload.


willkg:
This adds support for something the crash reporter will do in the future, but doesn't do, yet. Because of that, there wasn't a real thing to test against so it's possible we'll need to adjust this in the future.

self-r+

PR #303, opened 2019-03-19
Update docs regarding RabbitMQ -> Pub/Sub

willkg:


willkg:
These are all docs changes, so I'm not going to wait for CI. self-r+

PR #304, opened 2019-03-20
bug 1527343: switch to use CRASHPUBLISH_SERVICE_ACCOUNT_FILE

willkg:


willkg:
This makes setting the creds file consistent between Antenna and Socorro. That's a good thing. self-r+

PR #306, opened 2019-03-22
bug 1538202: remove preload

willkg:
Gunicorn can preload the entire application before spinning off workers
which saves time and memory. However, it calls gevent monkeypatch after
spinning off the workers, so a bunch of stuff has imported things that
need to be monkeypatched before monkeypatching. I'm pretty sure this
is what's causing the sporadic SystemExits we're seeing.

This shuts off preload.


willkg:
self-r+

PR #307, opened 2019-03-22
bug 1538202: add monkey.patch_all() calls to guarantee patching

willkg:
This adds a couple of `monkey.patch_all()` calls to guarantee gevent patching
before importing/usage of those things. This should fix issues in the pubsub
code. It also quells a warning when running pytest.


willkg:
self-r+

PR #308, opened 2019-03-25
bug 1538202: fix SystemExit (attempt 3)

willkg:
This adds calling of the (as far as I can tell) undocumented `init_gevent()` in grpc. This also downgrades grpcio to 0.18.0 which doesn't segfault when used with gunicorn and gevent.

That gets everything working in my local dev environment.

Having said that, this issue is sporadic and non-deterministic, so we'll have to land this and then wait a few days and see what happens.


antenna/ext/pubsub/crashpublish.py
```diff
@@ -2,11 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# The pubsub code needs monkeypatching to happen before it's imported.
-# This guarantees that.
-from gevent import monkey
-monkey.patch_all()  # noqa
```
willkg: This was an errant fix. The theory was that somehow the monkey patching wasn't happening before we imported pubsub things. That turned out not to be the issue but rather that the monkey patching didn't affect grpc which is not a pure Python library. Thus this fix had no effect.

antenna/gunicornworker.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class GeventGrpcWorker(gunicorn.workers.ggevent.GeventWorker):
+    """Gevent worker that also patches grpc."""
+    def patch(self):
+        super().patch()
+        grpc_gevent.init_gevent()
```
willkg: The `GeventWorker` already runs gevent patching. It needs to run the grpc `init_event`, too.

bin/run_web.sh
```diff
@@ -24,11 +24,10 @@ set -e
 
 # Launch the web-app
 gunicorn \
-    --preload \
```
willkg: We can't use preload by default anymore because it messes up the order of monkey patching. Fun.

requirements/constraints.txt
```diff
@@ -191,39 +191,35 @@ google-api-core==1.8.0 \
...
-    --hash=sha256:bcb44cd53beccc92c730254ad3d50715b67a7432e693961b566d982f759b1787 \
-    --hash=sha256:be1cbb6cad1d4242e3aaa4143eabcfbf383358f6c8e9951be2c497b65561b075 \
-    --hash=sha256:c4e38326fcab5c52fd1a8c8e0f908bfe830629a5ffc60793ec5545ef913d62d2 \
-    --hash=sha256:d03c0524d5953568f74269e0faebb1e880ba9f36ca8c773be397087c35bd8188 \
-    --hash=sha256:ea897ffa80276565acdd92349ef82a768db0e3327aacd4aec82f79ca10989689 \
-    --hash=sha256:edc50e8bcd10b165f34c3cf3e1d4f97e9c71b165b85a85b91cf3444000a17692 \
-    --hash=sha256:f96a2e97df522b50da9cb3795f08199b110ceab4146bf70ea7f6a3a0213786cc \
-    --hash=sha256:fadb649a69e3b08e01f090c24f0c8cccc122e92c362c1a1727b695a63be8416b \
-    --hash=sha256:fbe4360ff1689a9753cbf1b27dad11e683d39117a32a64372a7c95c6abc81b81
+grpcio==1.18.0 \
```
willkg: 1.19.0 causes segfaults now that we're (correctly) calling `init_gevent()`. Downgrading to 0.18.0 works fine.

I don't see anything in 1.19.0 fixes that might affect us, but who knows.

tests/unittest/conftest.py
```diff
@@ -2,11 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# Monkey patch stdlib with gevent-friendly bits--need to do this here
-# before we import anything else.
-from gevent import monkey
-monkey.patch_all()  # noqa
```
willkg: We don't need this anymore.

willkg:
I want to see how this goes. If this works out, then I don't think we *need* to do a rewrite *right now*, but I'm still not enthused about this situation. If this doesn't work out, on to a rewrite.

self-r+

PR #310, opened 2019-03-26
bug 1538202: fix the docker CMD line to use the new worker

willkg:
This fixes the CMD so that it's using the worker class that initializes
grpc for gevent.

Previously this was configurable, but since we require this class in order to
make gevent and grpc, I changed it so that it's no longer configurable.


willkg:
self-r+

PR #315, opened 2019-04-30
bug 1547804: process all ProcessType=gpu crash reports

willkg:
The thinking is that we don't get many of these crash reports so we should
process all the ones we do get. I'm estimating this increase processed
crashes per week by about 10,000 which should be fine.


willkg:
self-r+

PR #316, opened 2019-06-12
bug 1557706: specify why crash report was discarded

willkg:
This changes the breakpad crash report handling code to return something
like "Discarded=somereason" instead of "Discarded=1". Reasons are
alphanumeric one-word values that indicate what the problem was.

While doing that, I tweaked how extract_payload works for error
situations--now it throws a MalformedCrashReport exception.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -45,6 +45,15 @@ STATE_SAVE = 'save'
 STATE_PUBLISH = 'publish'
 
 
+class MalformedCrashReport(Exception):
+    """Exception raised when the crash report payload is malformed.
+
+    Message should be an alpha-numeric error code with no spaces.
+
+    """
+    pass
+
+
 class CrashReport:
     """Crash report structure."""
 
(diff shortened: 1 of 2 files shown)


antenna/breakpad_resource.py
```diff
@@ -225,16 +236,17 @@ def extract_payload(self, req):
...
                 mymetrics.incr('malformed', tags=['reason:wrong_content_type'])
+                raise MalformedCrashReport('wrongcontenttype')
             else:
                 mymetrics.incr('malformed', tags=['reason:no_boundary'])
-            return {}, {}
+                raise MalformedCrashReport('noboundary')
 
         content_length = req.content_length or 0
 
         # If there's no content, return an empty crash
```
willkg: Bah--I need to update these comments.

antenna/breakpad_resource.py
```diff
@@ -204,37 +213,40 @@ def extract_payload(self, req):
...
+                raise MalformedCrashReport('noboundary')
 
         content_length = req.content_length or 0
 
-        # If there's no content, return an empty crash
+        # If there's no content, raise MalformedCrashReport
         if content_length == 0:
             mymetrics.incr('malformed', tags=['reason:no_content_length'])
-            return {}, {}
+            raise MalformedCrashReport('nocontentlength')
```
willkg: I'm going to move the metrics to the handler.

willkg:
self-r+

PR #347, opened 2019-06-19
bug 1560168: add logging for json-and-kv situation

willkg:
We've got this problem where crash payloads with a JSON serialized extra field
are getting rejected by Antenna for unclear reasons. It seems like there are
extra fields involved, but we can't figure out where they're coming from or
what they look like.

This adds some logging that I can check later that'll help suss out the
problem.


willkg:
self-r+

PR #348, opened 2019-06-21
bug 1560168: ignore querystring params in HTTP POSTs

willkg:
The HTTP POST payload for a crash report is complete and has all the
information we need in it. Because of this, we should ignore the querystring
parameters which FieldStorage emits as items with no content type.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -285,26 +285,24 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
 
             mymetrics.histogram('crash_size', value=content_length, tags=['payload:uncompressed'])
 
-        fs = cgi.FieldStorage(fp=data, environ=req.env, keep_blank_values=1)
+        # Stomp on querystring so we don't pull it in
+        request_env = dict(req.env)
+        request_env['QUERY_STRING'] = ''
 
-        # NOTE(willkg): In the original collector, this returned request
-        # querystring data as well as request body data, but we're not doing
-        # that because the query string just duplicates data in the payload.
+        fs = cgi.FieldStorage(fp=data, environ=request_env, keep_blank_values=1)
 
         raw_crash = {}
         dumps = {}
 
-        field_info = []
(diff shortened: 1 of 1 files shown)


antenna/breakpad_resource.py
```diff
@@ -287,49 +287,44 @@ def extract_payload(self, req):
 
         fs = cgi.FieldStorage(fp=data, environ=req.env, keep_blank_values=1)
 
-        # NOTE(willkg): In the original collector, this returned request
-        # querystring data as well as request body data, but we're not doing
-        # that because the query string just duplicates data in the payload.
-
```
willkg: This note was here because the old collector both iterated over FieldStorage *and* also added stuff from querystring. However, iterating over FieldStorage picks up both HTTP POST and querystring parameters, so this comment isn't helpful and the old code was probably not right.

antenna/breakpad_resource.py
```diff
@@ -287,49 +287,44 @@ def extract_payload(self, req):
...
         has_kvpairs = False
 
         for fs_item in fs.list:
-            # NOTE(willkg): We saw some crashes come in where the raw crash ends up with
-            # a None as a key. Make sure we can't end up with non-strings as keys.
-            item_name = fs_item.name or ''
-            field_info.append('%s: %s' % (item_name, fs_item.type))
+            # If the field has no name, then it's probably junk, so let's drop it.
+            # If the field has no content-type, then it's from the querystring and
+            # it's redundant to the HTTP POST payload, so we can skip it.
```
willkg: Yay for better less hand-wavey comments!

willkg:
Turns out that the theory I could ignore fields that didn't have a content type was wrong. I tested it with the Firefox crash reporter. Ah well.

If I nix `QUERY_STRING` from the request environment, then `FieldStorage` doesn't pull querystring params. So that's the way we're going to fix this.

willkg:
I think this is the right thing. I can reproduce this in stage, so I can test it there.

PR #349, opened 2019-06-26
Update dependencies

willkg:
Note that this doesn't update:

* boto3 and botocore and friends since our s3mock doesn't work with newer versions.
* falcon since the next version includes significant API changes


willkg:
self-r+

PR #352, opened 2019-07-01
Update pytest and bandit to current versions

willkg:
This replaces pr #350 .

This doesn't update boto3 and friends because that breaks the s3 mock. Bug 1517807 covers that.

This doesn't update falcon because the jump to 2.0 involves some non-trivial reworking of things.


willkg:
These are dev-environment-only libs. self-r+

PR #353, opened 2019-07-23
bug 1560983: add black scaffolding and reformat the code

willkg:
This adds some minimal scaffolding for running `black`. It also reformats all the code using it.

Going forward, `black --check` will now get run during the lint step and affects whether PRs can get merged.


Makefile
```diff
@@ -32,6 +32,7 @@ default:
 	@echo "shell            - open a shell in the base container"
 	@echo "clean            - remove all build, test, coverage and Python artifacts"
 	@echo "lint             - check style with flake8"
+	@echo "black            - run black on black-formatted Python files"
```
willkg: I was thinking about these two make rules. It'd be nice to be the same across projects. It'd be nice to take into account that we're linting and fixing multiple languages. Tecken has "lintcheck" and "lintfix".

`mach` uses `mach lint` and `mach lint --fix`. We could do something like that, but it's tricky to pass arguments with make rules, so we'd still need a separate make rule.

I think I'm going to change this to "lint" and "lintfix".

Makefile
```diff
@@ -88,7 +89,11 @@ clean:
 
 .PHONY: lint
 lint: my.env .docker-build
-	${DC} run --rm --no-deps base  /bin/bash ./docker/run_lint.sh
+	${DC} run --rm --no-deps base /bin/bash ./docker/run_lint.sh
+
+.PHONY: black
+lintfix: my.env .docker-build
+	${DC} run --rm --no-deps base /bin/bash ./docker/run_lint.sh --fix
```
willkg: "lint" and "lintfix" are tools-agnostic names that sort of match what Tecken is doing and what "mach" does. We'll be able to use these across all the projects.

setup.cfg
```diff
@@ -3,10 +3,12 @@ universal = 1
...
 [flake8]
 ignore =
+    # E203: Whitespace before ':'; doesn't work with black
+    E203,
     # E501: line too long
     E501,
-    # W504: line break after binary operator
-    W504,
+    # W503: line break before operator; this doesn't work with black
+    W503,
```
willkg: Previously, we did line breaks after operators. Black does this differently, thus the switch.

setup.cfg
```diff
@@ -3,10 +3,12 @@ universal = 1
 
 [flake8]
 ignore =
+    # E203: Whitespace before ':'; doesn't work with black
+    E203,
```
willkg: Black adds whitespace before `:` characters in slices. e.g. `foo = [len(something) :5]`.

tests/systemtest/test_20mblimit.py
```diff
@@ -51,50 +52,52 @@ def _test_crash_size(self, posturl, size, crash_generator):
...
-        ))
-    @pytest.mark.parametrize('size, status_code', [
-        # up to and including 20mb should get an HTTP 200
-        ((20 * 1024 * 1024) - 1, 200),
-        ((20 * 1024 * 1024), 200),
-
-        # past 20mb, so this should fail with an HTTP 413
-        ((20 * 1024 * 1024) + 1, 413)
-    ])
+            "Requires nginx which you probably do not have running " "via localhost"
```
willkg: This is goofy--will fix.

tests/systemtest/test_20mblimit.py
```diff
@@ -51,50 +52,52 @@ def _test_crash_size(self, posturl, size, crash_generator):
...
         assert result == status_code
 
     @pytest.mark.skipif(
-        bool(os.environ.get('NONGINX')),
+        bool(os.environ.get("NONGINX")),
         reason=(
-            'Requires nginx which you probably do not have running '
-            'via localhost'
-        ))
+            "Requires nginx which you probably do not have running " "via localhost"
```
willkg: This is goofy--will fix.

tests/systemtest/test_content_length.py
```diff
@@ -70,28 +70,26 @@ def test_content_length_20(self, posturl, crash_generator):
...
+        assert str(resp.read(), encoding="utf-8") == "Discarded=1"
 
     @pytest.mark.skipif(
-        bool(os.environ.get('NONGINX')),
-        reason=(
-            'Requires nginx which you probably do not have running '
-            'via localhost'
-        ))
+        bool(os.environ.get("NONGINX")),
+        reason=("Requires nginx which you probably do not have running via localhost"),
```
willkg: This is goofy--will fix.

willkg:
I read through all the changes black made and fixed some goofy things. None of them change the behavior of the code. `make lint` works, `make lintfix` works, tests pass.

One thing I noticed is that the systemtests no longer pass. I broke them when I made the changes to bug 1557706. I went through all the failing tests and made sure they weren't failing because of black changes. I reopened that and will fix those in a followup PR.

This took about 2 hours to do. I spent some of that time looking at how to configure black and also figuring out how to get black and flake8 to play nice. For configuration, black only supports `pyproject.toml`, so I decided to put the configuration in the `run_lint.sh` shell script and keep it there. For getting black and flake8 to play nice, I just adjusted the flake8 ignores to match what black is doing. No biggie.

PR #354, opened 2019-07-23
bug 1557706: fix systemtests to work with discarded reasons

willkg:


antenna/breakpad_resource.py
```diff
@@ -325,6 +325,9 @@ def extract_payload(self, req):
                 has_kvpairs = True
                 raw_crash[fs_item.name] = fs_item.value
 
+        if not raw_crash:
+            raise MalformedCrashReport("no_annotations")
```
willkg: This causes malformed crash reports which have no annotations to get dropped earlier. Otherwise, they go through the throttler and get dropped because it's an "unsupported product". Dropping it here with a "no annotations" reason reduces some of the work involved but also yields a better reason for someone testing crash reporting.

willkg:
self-r+

PR #356, opened 2019-07-25
Update release.py

willkg:
This updates `release.py` from the socorro-release repo. It looks like nothing happened, but that's because I made changes to this `release.py` when I ran black to reformat the code. I copied the changes back to the socorro-release repo, but need to pick up the SHA change. This does that.


willkg:
This isn't affected by tests at all, so I'm just going to merge it. self-r+

PR #359, opened 2019-08-06
Update release.py and codify configuration

willkg:
This allows us to create bugs with:

```
$ ./bin/release.py make-bug
```

and make tags with:

```
$ ./bin/release.py make-tag --with-bug=BUGID
```


willkg:
self-r+

PR #366, opened 2019-11-07
bug 1517807: rewrite s3mock to work with latest botocore

willkg:


.editorconfig
```diff
@@ -20,4 +20,4 @@ insert_final_newline = false
 indent_style = tab
 
 [*.py]
-max_line_length = 100
+max_line_length = 88
```
willkg: This mostly affects me. I don't think anyone else uses editorconfig.

testlib/s3mock.py
```diff
@@ -4,24 +4,19 @@
...
-import os
+import unittest
 
-# boto has a vendored version of requests--we want to mock that one.
-from botocore.vendored import requests
-from botocore.vendored.requests.adapters import BaseAdapter
-from botocore.vendored.requests.models import Response
-
-# Note: DOUBLE-VENDORED!
-from botocore.vendored.requests.packages.urllib3.response import HTTPResponse
```
willkg: I'm going to miss this DOUBLE-VENDORED comment. It was really funny.

testlib/s3mock.py
```diff
@@ -43,6 +38,17 @@ class ThouShaltNotPass(Exception):
...
 
+@dataclasses.dataclass
+class Request:
+    method: str
+    url: str
+    body: io.BytesIO
+    headers: dict
+    scheme: str
+    host: str
+    port: int
```
willkg: This is my first experience with data classes. I like them. It's super easy to do this and you get to skip a ton of boilerplate. Data classes are new in 3.7. We could also use `attrs` which is in the requirements already. That's <3.7 friendly. With antenna, it's not a big deal. With socorro, we have some code we "export" to other libraries that support a wider range of Python versions, so we might want to be more careful.

testlib/s3mock.py
```diff
@@ -72,77 +86,27 @@ def build_response(self, request):
...
+        if "content-length" not in headers:
+            headers["content-length"] = len(body)
 
-        response.connection = self
+        response.request = request
 
         return response
 
 
-class FakeAdapter(BaseAdapter):
```
willkg: This is how we "mocked" things in requests. With urllib3, we're just patching urlopen instead, so we don't need this anymore. I folded the necessary bits into `S3Mock`.

testlib/s3mock.py
```diff
@@ -202,58 +166,16 @@ def ln(part):
     return b"\n".join(output)
 
 
-class RecordingAdapterShim:
```
willkg: I sort of used the recording feature, but not much. I decided to nix it and replace it with some code that shows what should have happened at the THWARTING point. It's not quite the same, but I think it still solves all my use cases.

testlib/s3mock.py
```diff
@@ -202,58 +166,16 @@ def ln(part):
...
-
 class S3Mock:
     """Provide a configurable mock for Boto3's S3 bits.
 
-    Boto3 uses botocore which uses s3transfer which uses requests to do REST
-    API calls.
+    Boto3 uses botocore which uses s3transfer which uses urllib3 to do REST API calls.
 
-    ``S3Mock`` mocks requests by creating a fake adapter which allows it to
-    intercept all outgoing HTTP requests. This lets us do two things:
```
willkg: Hilariously, I said "two things" here which used to be wrong, but now it's right!

testlib/s3mock.py
```diff
@@ -307,60 +229,62 @@ class S3Mock:
...
 
-    Usage::
-
-        with S3Mock() as s3:
-            s3.record(filename='s3mock.log')
-
-            # Do stuff here
+    If ``S3Mock`` is thwarting you, you can tell it to ``run_on_error`` and it'll
+    execute the urlopen and tell you want the response was. It'll also tell you
+    what it expected. This helps debugging assertions on HTTP conversations.
```
willkg: I should show usage of this. Though it's not wildly hard to figure out.

testlib/s3mock.py
```diff
@@ -307,60 +229,68 @@ class S3Mock:
...
+        # framework inside test scaffolding is "tricky".
+        print("THWARTED SEND:\nHTTP Request:\n%s" % serialize_request(req))
+        if self.expected_conv:
+            step = self.expected_conv[0]
+            print("Expected:\n%s %s\n%s" % (step.method, step.url, step.body))
+        else:
+            print("Expected: nothing")
+
+        if self._patcher and self._run_on_error:
+            resp = self._patcher.get_original()[0](
```
willkg: This calls the original `urlopen` on the pool instance which will perform the HTTP request and tell you what you got.

willkg:
The dependency updates were long overdue. They affect server environments. Rewriting s3mock only affects tests. This gets the tests passing again and cleans up some things about the s3mock so it's smaller. self-r+

PR #367, opened 2019-11-11
Cleanup retry code

willkg:
This cleans up the readability of the retry code.

I wanted to do a cleanup pass before copying this code into Socorro to use there. I need it there to help clean up the boto crash storage.

### DIFF
--- a/antenna/util.py
+++ b/antenna/util.py
@@ -13,6 +13,7 @@ import time
 import uuid
 
 import isodate
+from more_itertools import peekable
 
 
 logger = logging.getLogger(__name__)
@@ -190,24 +191,16 @@ def sanitize_dump_name(val):
 
 
 class MaxAttemptsError(Exception):
-    """Maximum attempts error.
+    """Maximum attempts error."""
 
-    .. :py:attribute:: msg
-       message for the exception
-
-    .. :py:attribute:: return_value
-       The last return value for the function.
-
-    """
-
-    def __init__(self, msg, ret):
+    def __init__(self, msg, ret=None):
         super().__init__(msg)
         self.return_value = ret
 
 
 def wait_time_generator():
     """Return generator for wait times."""
-    for amt in [1, 1, 5, 10, 30]:
+    for amt in [1, 1, 5, 10, 10]:
         yield amt
 
 
@@ -218,7 +211,7 @@ def retry(
     sleep_function=time.sleep,
     module_logger=None,
 ):
-    """Retry decorated function with wait times, max attempts and logging.
+    """Retry decorated function with wait times, max attempts, and logging.
 
     Example with defaults::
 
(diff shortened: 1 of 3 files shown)


antenna/util.py
```diff
@@ -190,24 +191,16 @@ def sanitize_dump_name(val):
...
-    def __init__(self, msg, ret):
+    def __init__(self, msg, ret=None):
         super().__init__(msg)
         self.return_value = ret
 
 
 def wait_time_generator():
     """Return generator for wait times."""
-    for amt in [1, 1, 5, 10, 30]:
+    for amt in [1, 1, 5, 10, 10]:
```
willkg: Changed the last item from 30 to 10 because waiting 30 seconds for anything seems unreasonable as a default.

I'm wondering whether a better default would be something like `[1, 1, 1, 1, 1]` or `[2, 2, 2, 2, 2]`.

antenna/util.py
```diff
@@ -293,33 +288,36 @@ def _retry_fun(*args, **kwargs):
...
-                            "%s: exception %s, retry attempt %s",
-                            fun.__qualname__,
-                            exc,
-                            attempts,
-                        )
-
-                    # If last attempt, re-raise the exception thrown
-                    if next_wait is None:
+                    # If it's a MaxAttemptsError, re-raise that
+                    if isinstance(exc, MaxAttemptsError):
```
willkg: Funny story. So if line 305 in the original code raises a `MaxAttemptsError` and the caller didn't specify which exceptions to retry on, then it would retry on `MaxAttemptsError` from a retryable return. That's definitely not what we want.

tests/unittest/test_util.py
```diff
@@ -107,6 +107,16 @@ def test_sanitize_dump_name(data, expected):
...
 
 
+def make_fake_sleep():
+    sleeps = []
+
+    def _fake_sleep(attempt):
+        sleeps.append(attempt)
+
+    _fake_sleep.sleeps = sleeps
+    return _fake_sleep
```
willkg: This closes over `sleeps` rather than having it as a variable in the outer environment. This is better and less error prone.

tests/unittest/test_util.py
```diff
@@ -121,56 +131,55 @@ def some_thing():
...
 
-        # This will fail on the first attempt because Exception is not
-        # in the list of retryable exceptions.
-        @retry(retryable_exceptions=ValueError, sleep_function=fake_sleep)
+        @retry(retryable_exceptions=ValueError, sleep_function=make_fake_sleep)
         def some_thing():
-            raise Exception
+            raise MyException
 
-        with pytest.raises(Exception):
```
willkg: `MaxAttemptsError` is a subclass of `Exception` so fixing this to throw and check for `MyException` removes the possibility that it's a `MaxAttemptsError` which should be a test failure.

willkg:
I think the yak-shaving on this is over.

willkg:
self-r+

PR #369, opened 2019-11-23
bug 1598780: convert cli to click

willkg:
This converts the two command line utilities to click. While doing that, I generalized both utilities so they use command line arguments rather than pulling from the environment and added a `make setup` that mirrors Socorro's.

To test:

1. run `make setup` which will delete and recreate the S3 bucket and then delete and recreate the Pub/Sub topic and subscription
2. run `make test`
3. run `make run` in one window and `make systemtest` in another which verifies that `make setup` did what it needed to do
4. run `make shell` and then `./bin/s3_cli.py list_objects antennabucket` which should list the objects in the antennabucket bucket created from the system test


willkg:
These utilities are only used in the local dev environment so there's no risk to server stuff. Between `make setup`, running the tests, and `make systemtest`, they're well-enough-tested. self-r+

PR #372, opened 2019-12-06
Tests fix

willkg:
The tests were doing this weird configuration thing where it wasn't pulling from the `local_dev.env` file, but instead ignoring local dev configuration.
    
As we adjust services around, it's easier to have the test environment configured from a file in one place rather than across multiple files.
    
This adjusts the test code to pull from the os environment, adds a `test.env` file, fixes some tests that were curiously broken, and then changes some formatting.

This only affects tests, so if they pass in the local dev environment (they do) and pass in CI (we'll see), then we're good here.

### DIFF
--- a/tests/unittest/conftest.py
+++ b/tests/unittest/conftest.py
@@ -8,7 +8,7 @@ import logging
 import sys
 from unittest import mock
 
-from everett.manager import ConfigManager
+from everett.manager import ConfigManager, ConfigDictEnv, ConfigOSEnv
 from falcon.request import Request
 from falcon.testing.helpers import create_environ
 from falcon.testing.client import TestClient
@@ -31,7 +31,7 @@ def pytest_runtest_setup():
     # Make sure we set up logging and metrics to sane default values.
     setup_logging(
         ConfigManager.from_dict(
-            {"HOST_ID": "", "LOGGING_LEVEL": "DEBUG", "LOCAL_DEV_ENV": "False"}
+            {"HOST_ID": "", "LOGGING_LEVEL": "DEBUG", "LOCAL_DEV_ENV": "True"}
         )
     )
     markus.configure([{"class": "markus.backends.logging.LoggingMetrics"}])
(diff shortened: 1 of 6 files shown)


tests/unittest/conftest.py
```diff
@@ -31,7 +31,7 @@ def pytest_runtest_setup():
     # Make sure we set up logging and metrics to sane default values.
     setup_logging(
         ConfigManager.from_dict(
-            {"HOST_ID": "", "LOGGING_LEVEL": "DEBUG", "LOCAL_DEV_ENV": "False"}
+            {"HOST_ID": "", "LOGGING_LEVEL": "DEBUG", "LOCAL_DEV_ENV": "True"}
```
willkg: Changing this to True gets rid of all the mozlog formatted logging. Blech.

tests/unittest/conftest.py
```diff
@@ -60,6 +60,15 @@ def _request_generator(method, path, query_string=None, headers=None, body=None)
...
     """Test client to ease testing with Antenna API"""
 
+    @classmethod
+    def build_config(cls, new_config=None):
+        """Build ConfigManager using environment and overrides."""
+        new_config = new_config or {}
+        config_manager = ConfigManager(
+            environments=[ConfigDictEnv(new_config), ConfigOSEnv(),]
+        )
+        return config_manager
```
willkg: This pulls from the environment and allows tests to override settings. Previously, it didn't pull from the environment at all.

tests/unittest/test_s3_crashstorage.py
```diff
@@ -130,6 +134,7 @@ def test_region_and_bucket_with_periods(
         client.rebuild_app(
             {
                 "CRASHSTORAGE_CLASS": "antenna.ext.s3.crashstorage.S3CrashStorage",
+                "CRASHSTORAGE_ENDPOINT_URL": "http://fakes3:4569",
```
willkg: Missing this caused this test to fail. This was a mystery I bumped into a while back, but couldn't figure out. What's going on is that since `CRASHSTORAGE_ENDPOINT_URL` isn't defined by default, then this would pick up AWS urls. Then we changed it to use the environment for configuration, so then it was an empty string and now it's set to this.

willkg:
I extracted this from my SQS PR. I needed to change some things to make the SQS testing easier.

willkg:
self-r+

PR #373, opened 2019-12-06
bug 1601455: add AWS SQS crashpublish support

willkg:
This implements an `SQSCrashPublish` class which publishes crash ids to an AWS SQS standard queue. It includes configuration, tests, and documentation for setting it up.

It is not enabled by default--the default crash publisher is still `NoOpCrashPublish`.

This won't affect production or the local dev environment. This adds the collector bits so that we can then add the processor/webapp bits.

### DIFF
--- a/antenna/ext/sqs/crashpublish.py
+++ b/antenna/ext/sqs/crashpublish.py
@@ -1,0 +1,169 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import logging
+import random
+
+import boto3
+from botocore.client import ClientError
+from everett.component import ConfigOptions
+import gevent
+
+from antenna.ext.crashpublish_base import CrashPublishBase
+from antenna.heartbeat import register_for_verification
+from antenna.util import retry
+
+
+logger = logging.getLogger(__name__)
+
+
+def wait_times_connect():
+    """Return generator for wait times with jitter between failed connection attempts."""
+    for i in [5] * 5:
+        yield i + random.uniform(-2, 2)  # nosec
+
+
+class SQSCrashPublish(CrashPublishBase):
+    """Publisher to AWS SQS.
+
+    **Required AWS SQS things**
+
+    When configuring credentials for this crashpublish object, you can do one of two
+    things:
+
+    1. provide ``ACCESS_KEY`` and ``SECRET_ACCESS_KEY`` in the configuration, OR
+    2. use one of the other methods described in the boto3 docs
+       http://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials
+
+    You also need to create an AWS SQS standard queue with the following settings:
+
+    ==========================  =========
+    Setting                     Value
+    ==========================  =========
+    Default Visibility Timeout  5 minutes
+    Message Retention Period    *default*
+    Maximum Message Size        *default*
+    Delivery Delay              *default*
+    Receive Message Wait Time   *default*
+    ==========================  =========
+
+    The AWS credentials that Antenna is configured with must have the following
+    Amazon SQS permissions on the SQS queue you created:
+
+    * ``sqs:GetQueueUrl``
+
+      Antenna needs to convert a queue name to a queue url. This requires the
+      ``sqs:GetQueueUrl``
+
+    * ``sqs:SendMessage``
+
+      Antenna sends messages to a queue--this is how it publishes crash ids.
+      This requires the ``sqs:SendMessage`` permission.
+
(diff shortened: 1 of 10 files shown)


antenna/ext/sqs/crashpublish.py
```diff
@@ -0,0 +1,169 @@
...
+    required_config.add_option(
+        "region",
+        default="us-west-2",
+        alternate_keys=["root:s3_region"],
+        doc="AWS region to connect to. For example, ``us-west-2``",
+    )
+    required_config.add_option(
+        "endpoint_url",
+        default="",
+        alternate_keys=["root:s3_endpoint_url"],
```
willkg: I forgot about the root thing. I'll fix the settings in the `local_dev.env`.

willkg: Actually, I'll leave them as is. It's easier.

antenna/ext/sqs/crashpublish.py
```diff
@@ -0,0 +1,169 @@
...
+        """Check AWS SQS connection health."""
+        try:
+            self.client.get_queue_url(QueueName=self.queue_name)
+        except Exception as exc:
+            state.add_error("SQSCrashPublish", repr(exc))
+
+    def publish_crash(self, crash_report):
+        """Publish a crash id to an AWS SQS queue."""
+        crash_id = crash_report.crash_id
+        self.client.send_message(QueueUrl=self.queue_url, MessageBody=crash_id)
```
willkg: The message body can be unformatted text, xml, or JSON.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.send_message

This is different than pubsub where we had to send bytes.

tests/unittest/test_pubsub_crashpublish.py
```diff
@@ -90,7 +90,6 @@ def test_verify_topic_no_topic(self, client, pubsub):
         with pytest.raises(NotFound):
             client.rebuild_app(
                 {
-                    "LOCAL_DEV_ENV": "True",
```
willkg: This is already specified in the `local_dev.env` file, so we don't need it here, too.

bin/sqs_cli.py
```diff
@@ -122,6 +122,11 @@ def list_queues(ctx):
 @click.pass_context
 def create(ctx, queue):
     """Create SQS queue."""
+    queue = queue.strip()
+    if not queue:
+        click.echo("Queue name required.")
+        return
+
```
willkg: This prevents the queue name from being the empty string which ... makes things interesting.

docker/config/local_dev.env
```diff
@@ -27,7 +27,7 @@ CRASHSTORAGE_BUCKET_NAME=antennabucket
 # Pub/Sub things for local development
 CRASHPUBLISH_SERVICE_ACCOUNT_FILE=
 CRASHPUBLISH_PROJECT_ID=local_dev_socorro
-CRASHPUBLISH_TOPIC_NAME=local_dev_socorro_normal
+CRASHPUBLISH_TOPIC_NAME=local_dev_socorro_standard
```
willkg: The name we're using in the documentation is "standard". Before we switched to pubsub, it was sometimes called "standard" and sometimes called "normal". Figure it's easier if I remove the "normal" case everywhere so it's easier to compare pubsub to sqs.

docker/config/local_dev.env
```diff
@@ -38,3 +38,4 @@ CRASHPUBLISH_REGION=us-east-1
 CRASHPUBLISH_ACCESS_KEY=foo
 CRASHPUBLISH_SECRET_ACCESS_KEY=foo
 CRASHPUBLISH_ENDPOINT_URL=http://localstack-sqs:4576
+CRASHPUBLISH_QUEUE_NAME=local_dev_socorro_standard
```
willkg: Need to set the queue name in the local dev env so the stuff works.

willkg:
This looks good. It's not active in a server environment, so even if there are bugs, nothing will be affected at this point.

I did some local testing by setting this in `my.env:

```
CRASHPUBLISH_CLASS=antenna.ext.sqs.crashpublish.SQSCrashPublish
```

Then I can post crashes to my local Antenna and see the queue using the `sqs_cli.py`:

```
$ make shell
app@931ad2446dab:/app$ ./bin/sqs_cli.py list_messages local_dev_socorro_standard
test
1885fae9-0275-4d93-95c7-058e20191206
app@931ad2446dab:/app$ 
```

The "test" comes from verifying that the queue can be published to when Antenna starts up. Socorro should already handle "test" since the PubSub crash publishing code does it, too.

After I've got the Socorro changes done, I can do more integration testing and make any needed fixes in Antenna then, but this seems like a good pass now.

self-r+

willkg:
Oh, whoops! I'm not working today. No one saw anything.

PR #375, opened 2020-01-03
bug 1606763: switch to dependabot

willkg:


willkg:
I think to set it up, we have to add a config file and then do something on the dependabot site. I don't understand that part, yet. I asked Greg for help.

willkg:
I talked with Greg. I'm going to land this and see if I can get dependabot working with Antenna.

.dependabot/config.yml
```diff
@@ -0,0 +1,9 @@
+version: 1
+update_configs:
+  - package_manager: "python"
+    directory: "/"
+    update_schedule: "monthly"
+
+  - package_manager: "docker"
+    directory: "/"
+    update_schedule: "monthly"
```
willkg: We want dependabot to look at Dockerfiles. I'm pretty sure this will work. If not, hopefully we find out. This saves us from fiddling with minor Python version updates.

.dependabot/config.yml
```diff
@@ -0,0 +1,9 @@
+version: 1
+update_configs:
+  - package_manager: "python"
+    directory: "/"
+    update_schedule: "monthly"
```
willkg: For Antenna, this is the big one we want help with. This will keep Python dependencies up to date. The one we don't want to update (for now) is Falcon. I'll add a thing later to prevent dependabot from updating that.

willkg:
self-r+

PR #417, opened 2020-01-07
Add urlwait calls to docker/run_setup.sh

willkg:
`docker/run_setup.sh` requires services to be started. If the machine is fast enough, they may not be running by the time `docker/run_setup.sh` is called which causes it to fail. This reduces that by adding urlwait calls.


willkg:
self-r+

PR #418, opened 2020-01-08
Update test data container

willkg:
We just updated the Python docker container. This updates the test data container to be the same.


willkg:
self-r+

PR #419, opened 2020-01-09
bug 1608104: add note about annotations in crash report

willkg:
This adds a note to the raw crash that tells us how the annotations were encoded in the crash report. Were they in a JSON-encoded value in the "extra" field? Were they encoded as key/value pairs in the
multipart/form-data?


willkg:
self-r+

PR #420, opened 2020-01-13
bug 1604848: reject incoming thunderbird > 68 and seamonkey > 2.49.5

willkg:
This adds two throttling rules. One rejects incoming Thunderbird crash reports for major version > 68. The other rejects incoming SeaMonkey crash reports > 2.49.5.

This adds notes to remove these rules and remove Thunderbird and SeaMonkey from the supported products in January 2021. It's possible the timeline may change between now and then. If that happens, someone will update the notes.

### DIFF
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -230,6 +230,49 @@ def match_b2g(throttler, data):
     return False
 
 
+def match_seamonkey_gt_2_49_5(throttler, data):
+    """Reject SeaMonkey > 2.49.5 (bug #1604848).
+
+    NOTE(willkg): Remove this January 2021.
+
+    """
+    product = data.get("ProductName", "").lower().strip()
+    version = data.get("Version", "")
+
+    if product != "seamonkey" or not version:
+        return False
+
+    try:
+        version = tuple(int(n) for n in version.split("."))
+    except ValueError:
+        return False
+
+    if version > (2, 49, 5):
+        return True
+
+    return False
+
+
+def match_thunderbird_gt_68(throttler, data):
+    """Reject Thunderbird major version > 68 (bug #1604848).
+
+    NOTE(willkg): Remove this January 2021.
+
+    """
+    product = data.get("ProductName", "").lower().strip()
+    version = data.get("Version", "")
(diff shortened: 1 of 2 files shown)


willkg:
@jwhitlock Can you double-check the logic? I think it's right, but I did a few permutations of it and I'm human. Definitely warrants a second set of eyes.

willkg:
For the record, the current version of SeaMonkey is 2.49.5: https://www.seamonkey-project.org/releases/

They don't do releases often, so they may not even do a release by August 2020. If that ends up being the case, we may drop support altogether at that point. It depends on conversations that happen between now and then.

Thunderbird is on 68.4.1. They do releases periodically and I think they track Firefox ESR. They should have a new Thunderbird out by then. If not, then the plan is to stop supporting Thunderbird altogether in August 2020.

Once this lands, I'll write up a bug for those changes which will be the codified plan and discussions can happen in that bug to adjust it as necessary.

willkg:
On second thought, I'll write up the bug now.

willkg:
^^^ That restricts the checks to the release channel only.

antenna/throttler.py
```diff
@@ -230,6 +230,51 @@ def match_b2g(throttler, data):
...
+    """
+    product = data.get("ProductName", "").lower().strip()
+    version = data.get("Version", "")
+    channel = data.get("ReleaseChannel", "")
+
+    if product != "seamonkey" or not version or channel != "release":
+        return False
+
+    try:
+        version = tuple(int(n) for n in version.split("."))
```
jwhitlock: I'm not sure what happens when the version number is something like `2.70a1`, but I'm guessing `ReleaseChannel` is set to something other than `"release"`.

willkg: I don't think there's a case where the release channel is "release" and the version has non-digits in it. However, we get junk data periodically, so if we had a junk version, it'd raise a `ValueError` and that'll cause this function to return `False` which is the right thing to do.

That variation should be covered by a test, but it's not. I'll add that now.

willkg:
Thank you!

PR #422, opened 2020-01-21
bug 1610540: add code for removing fields

willkg:
This adds code for removing fields that should never exist in crash reports before the crash reports are saved anywhere.

Currently, this removes `TelemetryClientId` and `TelemetryServerURL`. These are stripped from the crash report before being sent, but in case they aren't, this removes them in the collector.

@jwhitlock wondered why we don't have a schema for what gets accepted and reject anything that's not in the schema. That wasn't doable before, but might be doable now. It's a pretty big change from how we do things currently. It might mean we need a schema per product. It's something we should think about at some point in 2020 or 2021.


willkg:
self-r+

PR #446, opened 2020-02-18
bug 1614985: handle JSONDecodeError in json value

willkg:
If the JSON value kicks up a `JSONDecodeError`, then there's something wrong with the JSON and Antenna should reject the crash report with an error.

### DIFF
--- a/tests/unittest/test_breakpad_resource.py
+++ b/tests/unittest/test_breakpad_resource.py
@@ -103,7 +103,7 @@ class TestBreakpadSubmitterResource:
         )
 
         bsp = BreakpadSubmitterResource(self.empty_config)
-        with pytest.raises(MalformedCrashReport):
+        with pytest.raises(MalformedCrashReport, match="wrong_content_type"):
             bsp.extract_payload(req)
 
     def test_extract_payload_compressed(self, request_generator):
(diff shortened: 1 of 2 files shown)


tests/unittest/test_breakpad_resource.py
```diff
@@ -103,7 +103,7 @@ def test_extract_payload_bad_content_type(self, request_generator):
         )
 
         bsp = BreakpadSubmitterResource(self.empty_config)
-        with pytest.raises(MalformedCrashReport):
+        with pytest.raises(MalformedCrashReport, match="wrong_content_type"):
```
willkg: This improves the test to verify the specific malformed reason.

tests/unittest/test_breakpad_resource.py
```diff
@@ -151,6 +151,18 @@ def test_extract_payload_json(self, request_generator):
         expected_dumps = {"upload_file_minidump": b"abcd1234"}
         assert bsp.extract_payload(req) == (expected_raw_crash, expected_dumps)
 
+    def test_extract_payload_bad_json(self, request_generator):
+        # If the JSON doesn't parse (invalid control character), it raises
+        # a MalformedCrashReport
+        data, headers = multipart_encode({"extra": '{"ProductName":"Firefox\n"}'})
```
willkg: This has a `\n` but that's converted to a carriage return during Python parsing and carriage return is an invalid character in a JSON string.

willkg:
self-r+

PR #477, opened 2020-03-04
bug 1619101: update seamonkey filter to 2.53.1

willkg:
SeaMonkey 2.53.1 released over the weekend and the collector should accept crash reports from it for now. This updates the filter accordingly.


willkg:
self-r+

PR #481, opened 2020-03-11
Clean up documentation

willkg:
1. updates architecture docs to cover publishing step
2. replaces Pub/Sub with SQS
3. moves hand-wavey bits out of the breakpad section into appropriate
   sections in the spec and dev docs


willkg:
self-r+

PR #482, opened 2020-03-16
bug 1618201: fix gunicorn worker class to be configurable

willkg:
This was using the old gunicorn worker class that we created to handle making grpcio aware we were using gevent. Now that we're not using Pub/Sub, we aren't using grpcio and thus we don't need that anymore and can go back to the gevent worker.

The infrastructure sets the `GUNICORN_WORKER_CLASS` variable, so we should use that.


willkg:
This should fix stage deploys. self-r+

PR #490, opened 2020-03-24
bug 1619644: handle None values in throttler

willkg:
It's now possible for the collector to get `None` values for crash annotations. `None` is not a string which makes some of the throttler rules sad.

This adds a `safe_get` function which gets the value from the crash and forces it to be a string. Then we can use string functions on it.

### DIFF
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -28,6 +28,20 @@ FAKEACCEPT = 3  # return crashid as if we accepted, but throw away--USE CAUTION!
 RESULT_TO_TEXT = {0: "ACCEPT", 1: "DEFER", 2: "REJECT", 3: "FAKEACCEPT"}
 
 
+def safe_get(data, key, default=""):
+    """Return a sanitized data[key].
+
+    Throttle rule conditions operate on strings since strings are the predominant value
+    type for crash report annotations.
+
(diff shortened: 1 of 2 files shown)


antenna/throttler.py
```diff
@@ -28,6 +28,20 @@
...
+
+    Throttle rule conditions operate on strings since strings are the predominant value
+    type for crash report annotations.
+
+    This forces the value to be a string. For example, `None` will become `"None"`.
+
+    :return: string value
+
+    """
+    return str(data.get(key, default))
```
willkg: Having a single function that does this is easier than adjusting the code everywhere. This is a sanitizing function, so as we hit issues with new and exciting data problems, we can tweak this function accordingly.

antenna/throttler.py
```diff
@@ -28,6 +28,20 @@
 RESULT_TO_TEXT = {0: "ACCEPT", 1: "DEFER", 2: "REJECT", 3: "FAKEACCEPT"}
 
 
+def safe_get(data, key, default=""):
+    """Return data[key] or empty string.
```
willkg: Oops--this is wrong now. I need to update this.

willkg:
self-r+

PR #491, opened 2020-03-27
bug 1624949: throttle ShutDownKill crash reports

willkg:
These are not really *crashes* and we get an awful lot of them. This reduces the number we're getting by 90%.

This adds a new throttle result: `CONTINUE`. When `CONTINUE` is returned, the throttler should continue through the rules--it's not an end state.


willkg:
I tested this locally by sending a minimal crash report:

```
{"ProductName": "Firefox", "ipc_channel_error": "ShutDownKill", "ReleaseChannel": "nightly"}
```

using this:

```
python3 -m testlib.mini_poster --url "http://localhost:8000/submit" --raw_crash crashdata/minimal.json
```

Most crash reports were rejected (yay!) and some were accepted (yay!).

self-r+

PR #534, opened 2020-05-27
bug 1513020: reject incoming crash reports with build id > 2 years

willkg:
This reject all incoming crash reports where the build id indicates it's over 2 years old. We don't have symbols for builds this old, so these crash reports aren't actionable.

### DIFF
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -9,6 +9,7 @@ process), defer (save, but not process), and reject.
 
 """
 
+import datetime
 import importlib
 import logging
 import random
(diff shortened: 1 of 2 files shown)


antenna/throttler.py
```diff
@@ -287,6 +301,13 @@ def match_unsupported_product(throttler, data):
 
 #: Ruleset for Mozilla's crash collector throttler
 MOZILLA_RULES = [
+    # If it's got an old build id, reject it now
+    Rule(
+        rule_name="has_old_buildid",
+        key="*",
+        condition=match_old_buildid,
+        result=REJECT,
+    ),
```
willkg, line 310: Placing this first is interesting in that this might match all the `has_hangid_and_browser` crash reports and the `infobar_is_true` crash reports. If that's true, we will be able to remove those rules.

tests/unittest/test_throttler.py
```diff
@@ -204,6 +245,7 @@ def test_hangid(self, throttler):
         }
         assert throttler.throttle(raw_crash) == (REJECT, "has_hangid_and_browser", 100)
 
+    @freeze_time("2019-05-27 12:00:00", tz_offset=0)
```
willkg, line 248: The infobar test uses a build id, too. We want to make sure that test doesn't get the right answer for the wrong reason.

willkg:
self-r+

PR #598, opened 2020-08-24
bug 1660793: redo system tests

willkg:
This redoes the system tests so you can more easily run them against
different environments.

To test:

1. in another terminal, run `make run`
2. run against local dev
   ```
   $ make shell
   app@xxx:/app$ cd systemtests
   app@xxx:/app/systemtests$ ./test_env.sh local
   ...
   ```
   Verify it **does not** run the nginx tests, but **does** run the postcheck tests.
3. run against stage
   ```
   $ make shell
   app@xxx:/app$ cd systemtests
   app@xxx:/app/systemtests$ ./test_env.sh stage
   ...
   ```
   Verify it **does** run the nginx tests, but **does not** run the postcheck tests.

Note: some of the max size tests fail because we just changed the max size--I'll fix that in a followup PR.


willkg:
self-r+

PR #599, opened 2020-08-24
bug 1648387: increase max body size checks in systemtests

willkg:
We're changing the max body size to 25mb and this updates the
systemtests accordingly.


willkg:
self-r+

PR #600, opened 2020-08-24
bug 1381712: wrap fieldstorage call in try/except for bad boundaries

willkg:
If the parsed boundary is not a `str`, then `cgi.FieldStorage` throws a
`ValueError`. This catches that and throws a `MalformedCrashReport`.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -293,7 +293,12 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
         request_env = dict(req.env)
         request_env["QUERY_STRING"] = ""
 
-        fs = cgi.FieldStorage(fp=data, environ=request_env, keep_blank_values=1)
+        try:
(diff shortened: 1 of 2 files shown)


antenna/breakpad_resource.py
```diff
@@ -293,7 +293,12 @@ def extract_payload(self, req):
         request_env = dict(req.env)
         request_env["QUERY_STRING"] = ""
 
-        fs = cgi.FieldStorage(fp=data, environ=request_env, keep_blank_values=1)
+        try:
+            fs = cgi.FieldStorage(fp=data, environ=request_env, keep_blank_values=1)
+        except ValueError:
+            # cgi.FieldStorage throws a ValueError if the boundary is not a str;
+            # treat this as an invalid payload
+            raise MalformedCrashReport("malformed_boundary")
```
willkg, line 301: The existing testlib makes this a really hard test to write. I'm tinkering with one.

willkg:
self-r+

PR #601, opened 2020-08-24
bug 1619640: fix UnicodeDecodeError when decoding JSON

willkg:
If the JSON-encoded value has a bad character in it, it'll kick up a `UnicodeDecodeError`. This adds handling for that situation.


willkg:
self-r+

PR #635, opened 2020-10-01
Update dependabot configuration

willkg:
This updates the dependabot configuration to the GitHub-native version.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,9 +1,0 @@
-version: 1
-update_configs:
-  - package_manager: "python"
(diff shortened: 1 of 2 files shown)


willkg:
I did this for Socorro already and it seems like it worked. I'm opting for monthly because it's easier to bulk-update everything at the same time rather than have a daily series of issues trickling in.

@relud Does this look right? Are there things we should be doing differently dependabot-wise?

relud (approved):
lgtm

willkg:
Thank you!

PR #640, opened 2020-10-02
bug 1668583: redo test scaffolding

willkg:
This changes the base image to use an entrypoint which includes running
a shell and moves the previous cmd to docker/run_web.sh.

This redoes how tests work so they run in either a test or a test-ci
container, the latter of which doesn't volume mount . which CircleCI
can't do.


willkg:
self-r+

PR #643, opened 2020-10-06
Update docker version in CircleCI config

willkg:
This explicitly sets the Docker version. If we don't set the version,
CircleCI defaults to using 17.03.0 which is a version that CircleCI is
deprecating.


willkg:
self-r+

PR #657, opened 2020-12-01
bug 1680046: add payload_compressed annotation

willkg:
This annotation captures whether the incoming payload was compressed or
not at ingestion.


willkg:
self-r+

PR #669, opened 2021-01-21
bug #1687987: remove unhelpful added fields from raw crash

willkg:
Antenna adds a bunch of fields to the raw crash to help us understand what happened at ingestion. This removes the following:

* type_tag -- we don't use this
* legacy_processing -- this isn't helpful since we don't need to distinguish between ACCEPT and DEFER anymore
* throttle_rate -- this isn't interesting to know after the fact
* timestamp -- we can use submitted_timestamp to calculate how long it took to collect the crash

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -374,7 +374,6 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
         result, rule_name, throttle_rate = self.throttler.throttle(raw_crash)
 
         # Save the results in the raw_crash itself
-        raw_crash["legacy_processing"] = result
         raw_crash["throttle_rate"] = throttle_rate
 
         return result, rule_name, throttle_rate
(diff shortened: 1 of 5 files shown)


antenna/breakpad_resource.py
```diff
@@ -110,9 +116,6 @@ class BreakpadSubmitterResource(RequiredConfigMixin):
         default="upload_file_minidump",
         doc="The name of the field in the POST data for dumps.",
     )
-    required_config.add_option(
-        "dump_id_prefix", default="bp-", doc="The crash type prefix."
-    )
```
willkg, line 115: We've never set this, so I moved the `bp-` to where it's used.

antenna/breakpad_resource.py
```diff
@@ -376,10 +379,6 @@ def get_throttle_result(self, raw_crash):
         # throttle the crash.
         result, rule_name, throttle_rate = self.throttler.throttle(raw_crash)
 
-        # Save the results in the raw_crash itself
-        raw_crash["legacy_processing"] = result
-        raw_crash["throttle_rate"] = throttle_rate
-
         return result, rule_name, throttle_rate
```
willkg, line 382: This method doesn't do much other than call another thing. I thought about removing this interstitial method, but then decided to think about that later.

PR #670, opened 2021-01-27
bug 1689035: move scripts to bin/

willkg:
There were some scripts in `bin/` and some in `docker/` and there wasn't much rhyme or reason for it. This changes things so that all scripts are in `bin/`.

The backstory for this is that I've got a bunch of projects that are different in weird, unhelpful ways and I'm trying to make them more the same so they're easier to maintain.


bin/run_web.sh
```diff
@@ -2,28 +2,29 @@
...
 
-# For development, it probably makes sense to use one process (``--workers=1``)
-# that can handle multiple concurrent connections (``--worker-connections=4``).
-# The number of connections you want to handle simultaneously depends on your
-# setup and all that.
-#
-# Make sure you use the ``gevent`` worker class (``--worker-class=gevent``).
-# Otherwise it's not going to use the gevent WSGI app and then you're not going
-# to be able to handle multiple network connections concurrently.
+set -euxo pipefail
```
willkg, line 9: We want this to be `set -euxo pipefail` so it has tracing on and tells us the values of the environment variables.

willkg: Also, fun fact that since we had scripts in two places, we had to `run_web.sh` scripts. Oops.

PR #783, opened 2022-02-24
bug 1756223: change ci docker image

willkg:
This switches from mozilla/cidockerbases which is deprecated to
cimg/base which is not.


willkg:
I'm done yak-shaving for now. I need to update the branch protection rules in GitHub, but I'll do that after I land this.

I can't tell if docker layer caching is helping much, yet. We'll see after a few builds, I suppose.

PR #784, opened 2022-02-24
bug 1756223: fix lint and test CI jobs

willkg:
This undoes the "touch .docker-build" line. Docker image tags don't seem
to be available between jobs.

This also fixes the lint job to use the make rule so we're not
specifying it in two places.


willkg:
That was kind of painful. Let's see if this passes after being merged.

PR #785, opened 2022-02-24
bug 1756223: fix version.json

willkg:
The version.json file isn't in the image that got pushed to dockerhub.
This reworks things so that the build job saves the image to the
workspace and that's used for subsequent jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,10 +61,8 @@ jobs:
 
       - run:
           name: Verify requirements.txt file
-          # yamllint disable rule:line-length
           command: |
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -43,9 +43,6 @@ jobs:
             "$CIRCLE_BUILD_URL" > version.json
           # yamllint enable rule:line-length
 
-      - store_artifacts:
-          path: version.json
```
willkg, line 47: I don't know why we ever did this. We don't use this artifact anywhere as near as I can tell.

PR #789, opened 2022-02-25
bug 1748256: handle when extra is not a dict value

willkg:
extra should be a JSON-encoded dict of key/value pairs. If it's not a
dict, then that's malformed and we should treat it as such.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -206,6 +206,9 @@ class BreakpadSubmitterResource:
                     # of "malformed JSON" situations.
                     raise MalformedCrashReport("bad_json")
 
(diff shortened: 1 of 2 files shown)


antenna/breakpad_resource.py
```diff
@@ -206,6 +206,9 @@ def extract_payload(self, req):
                     # of "malformed JSON" situations.
                     raise MalformedCrashReport("bad_json")
 
+                if not isinstance(raw_crash, dict):
+                    raise MalformedCrashReport("bad_json")
```
willkg, line 210: We might want to enforce a schema here at some point, but I'm not ready to do that, yet.

PR #790, opened 2022-02-25
bug 1747398: add support for multipart/mixed

willkg:
This allows the payload to be either multipart/form-data or
multipart/mixed.


willkg:
After landing this, I think we still need to update the crash report payload spec docs in socorro.

PR #806, opened 2022-03-03
bug 1757854: make systemtests easier to use for debugging

willkg:
This moves some bits around in the systemtest code to make it easier to
use for debugging crash report submission bugs.

### DIFF
--- a/systemtest/conftest.py
+++ b/systemtest/conftest.py
@@ -16,9 +16,9 @@ import pytest
 REPO_ROOT = Path(__file__).parent.parent
 sys.path.insert(0, str(REPO_ROOT))
 
-# Set up logging to log at DEBUG level; this is quelled by pytest except when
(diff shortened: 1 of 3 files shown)


systemtest/conftest.py
```diff
@@ -16,9 +16,9 @@
 REPO_ROOT = Path(__file__).parent.parent
 sys.path.insert(0, str(REPO_ROOT))
 
-# Set up logging to log at DEBUG level; this is quelled by pytest except when
+# Set up logging to log at INFO level; this is quelled by pytest except when
 # there are errors
-logging.basicConfig(level=logging.DEBUG)
+logging.basicConfig(level=logging.INFO)
```
willkg, line 21: At DEBUG, any failure dumps unbelievably large amounts of output that isn't particularly interesting, so I bumped this to INFO.

PR #809, opened 2022-03-21
bug 1760187: improve s3_cli and extract_payload error messages

willkg:
This improves the s3_cli giving it the ability to download files.

This improves the error messages from extract_payload so it's easier to identify and fix crash reporter client errors.

Previously, we had:

* malformed_no_content_type
* malformed_wrong_content_type
* malformed_no_boundary
* malformed_no_content_length
* malformed_bad_gzip
* malformed_bad_json
* malformed_no_annotations
* malformed_has_json_and_kv

Now, we've got:

* malformed_no_content_type
* malformed_wrong_content_type
* malformed_no_boundary
* malformed_no_content_length
* malformed_bad_gzip
* malformed_invalid_json
* malformed_invalid_json_value
* malformed_invalid_annotation_value
* malformed_invalid_payload_structure
* malformed_no_annotations
* malformed_has_json_and_kv

That makes it easier to differentiate between "no annotations" because there are malformed parts or "no annotations" because there weren't any annotations. It also clarifies "invalid json" (there was a decoding error) from "invalid json value" (the value isn't a python dict).

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -48,8 +48,6 @@ class MalformedCrashReport(Exception):
 
     """
 
-    pass
-
 
 class BreakpadSubmitterResource:
     """Handles incoming breakpad-style crash reports.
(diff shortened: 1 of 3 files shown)


mkestler-mozilla (approved):

antenna/breakpad_resource.py
```diff
@@ -206,15 +204,22 @@ def extract_payload(self, req):
...
                     has_kvpairs = True
-                    raw_crash[part.name] = part.get_text()
+                    try:
+                        raw_crash[part.name] = part.get_text()
+                    except MultipartParseError as mpe:
+                        logger.error(
+                            f"extract payload text part exception: {mpe.description}"
+                        )
+                        msg = "invalid_annotation_value"
+                        raise MalformedCrashReport(msg) from mpe
```
willkg, line 222: I can probably fold this back up.

antenna/breakpad_resource.py
```diff
@@ -206,15 +204,22 @@ def extract_payload(self, req):
...
                         # one of the characters. The JSONDecodeError can happen in a variety
                         # of "malformed JSON" situations.
-                        raise MalformedCrashReport("bad_json")
+                        raise MalformedCrashReport("invalid_json")
 
                     if not isinstance(raw_crash, dict):
-                        raise MalformedCrashReport("bad_json")
+                        raise MalformedCrashReport("invalid_json_value")
 
                 elif part.content_type.startswith("text/plain"):
```
willkg, line 212: I should add a check to see if there's a filename. If there's a filename, then this can't be a text annotation.

PR #865, opened 2022-09-06
bug 1604845: remove entropy from raw crash path

willkg:
This removes the entropy portion from the raw crash path so it goes from this:

```
v2/raw_crash/ENTROPY/DATE/CRASHID
```
to this:
```
v1/raw_crash/DATE/CRASHID
```

That matches the work done in Socorro.

I ran the tests, posted some crash reports, and verified the contents of the bucket are in the right directory structure.

### DIFF
--- a/antenna/ext/s3/crashstorage.py
+++ b/antenna/ext/s3/crashstorage.py
@@ -29,9 +29,8 @@ class S3CrashStorage(CrashStorageBase):
                    <CRASHID>
            v2/
                raw_crash/
-                   <ENTROPY>/
-                       <YYYYMMDD>/
-                           <CRASHID>
+                   <YYYYMMDD>/
+                       <CRASHID>
 
     """
 
@@ -66,8 +65,7 @@ class S3CrashStorage(CrashStorageBase):
         self.connection.check_health(state)
 
     def _get_raw_crash_path(self, crash_id):
-        return "v2/raw_crash/{entropy}/{date}/{crash_id}".format(
-            entropy=crash_id[:3],
+        return "v2/raw_crash/{date}/{crash_id}".format(
             date=get_date_from_crash_id(crash_id),
             crash_id=crash_id,
         )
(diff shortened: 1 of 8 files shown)


antenna/ext/s3/crashstorage.py
```diff
@@ -66,8 +65,7 @@ def check_health(self, state):
         self.connection.check_health(state)
 
     def _get_raw_crash_path(self, crash_id):
-        return "v2/raw_crash/{entropy}/{date}/{crash_id}".format(
-            entropy=crash_id[:3],
+        return "v2/raw_crash/{date}/{crash_id}".format(
```
willkg, line 68: I just re-re-re-read the bug and noticed I intended to do a `v3`. But this is really just like v1, so we should just switch it back. Nothing else depends on the versioning of the file system.

systemtest/test_content_length.py
```diff
@@ -65,20 +65,23 @@ def test_content_length_0(self, posturl, crash_generator):
...
         # Add wrong content-length
         headers["Content-Length"] = "20"
 
         resp = http_post(posturl, headers, payload)
 
         assert resp.getcode() == 400
         assert (
-            str(resp.read(), encoding="utf-8") == "Discarded=malformed_no_annotations"
+            str(resp.read(), encoding="utf-8")
+            == "Discarded=malformed_invalid_payload_structure"
```
willkg, line 84: The discarded messages changed a while ago in 29b5a831e86f4df480ed3adc10ade93638485987 (March 2022). I must not have run the systemtests since then. :(

testlib/mini_poster.py
```diff
@@ -277,18 +277,17 @@ def cmdline(args):
...
         raw_crash_path = Path(parsed.raw_crash)
-        if str(raw_crash_path.parents[3]).endswith("v2"):
+        if str(raw_crash_path.parents[3]).endswith("v1"):
             logger.info("Trying to find dump_names and dumps...")
             crashid = str(Path(parsed.raw_crash).name)
 
-            # First, raw_crash is ROOT/v2/raw_crash/ENTROPY/DATE/CRASHID, so
-            # find the root.
-            root_path = Path(parsed.raw_crash).parents[4]
+            # First, raw_crash is ROOT/v2/raw_crash/DATE/CRASHID, so find the root.
```
willkg, line 289: This v2 should be a v1.

PR #887, opened 2022-11-16
bug 1800531: accept all phc crash reports

willkg:


antenna/throttler.py
```diff
@@ -212,7 +212,7 @@ def always_match(throttler, crash):
 def match_infobar_true(throttler, data):
     """Match crashes we need to filter out due to infobar bug.
 
-    Bug #1425949.
```
willkg, line 215: This bug number has been wrong for 5 years. :(

PR #889, opened 2022-11-16
bug 1791376: be more defensive about build id matching

willkg:


antenna/throttler.py
```diff
@@ -259,16 +259,23 @@ def match_unsupported_product(throttler, data):
     return False
 
 
+BUILDID_RE = re.compile(r"^20\d{12}$")
```
willkg, line 262: The first 8 characters are YYYYMMDD, so the first 2 will be 20 unless by some amazing miracle this software survives the next 80 years which I'm extremely skeptical of.

PR #902, opened 2023-01-03
bug 1791209: support docker compose v2+

willkg:
Previously, we supported docker-compose < v2. This changes it so that we require at least v2.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ version: 2.1
 jobs:
   main:
     docker:
-      - image: cimg/base:2022.02
+      - image: cimg/base:2023.01
         auth:
           username: $DOCKER_USER
           password: $DOCKER_PASS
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -61,21 +59,7 @@ jobs:
...
-            docker-compose run --rm --no-deps ci shell ./bin/run_verify_reqs.sh
-
-      - run:
-          name: Save image
-          # yamllint disable rule:line-length
-          command: |
-            mkdir -p workspace
-            docker save -o workspace/antenna-deploy-base.tar local/antenna_deploy_base:latest
-            gzip --fast workspace/antenna-deploy-base.tar
-          # yamllint enable rule:line-length
```
willkg, line 73: This bit is left over from when we did parallel jobs in CI. We don't need to do this anymore.

PR #935, opened 2023-09-14
bug 1849352: add throttler rule for Firefox ESR Windows <= 8.1

willkg:
This adds a throttler rule to accept only 25% of Firefox, ESR, Windows <= 8.1 crash reports.

### DIFF
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -11,6 +11,7 @@ process), defer (save, but not process), and reject.
 
 import datetime
 import importlib
+import json
 import logging
 import random
 import re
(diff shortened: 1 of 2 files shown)


antenna/throttler.py
```diff
@@ -281,6 +282,40 @@ def match_old_buildid(throttler, data):
     return buildid_date < (now - datetime.timedelta(days=730))
 
 
+WINDOWS_8_BUILD_NUMBER = 9600
```
willkg, line 285: This is wrongly named. It's Windows 8.1.

antenna/throttler.py
```diff
@@ -372,6 +407,19 @@ def match_old_buildid(throttler, data):
         condition=lambda throttler, x: x == "ShutDownKill",
         result=(10, CONTINUE, REJECT),
     ),
+    # Accept 25% crash reports from Firefox ESR Windows <= 8.1
+    Rule(
+        rule_name="is_unsupported_firefox_esr_windows_version",
+        key="*",
+        condition=(
+            lambda throttler, data: (
+                safe_get(data, "Product") == "esr"
```
willkg, line 416: This line is wrong. Plus this needs a test.

antenna/throttler.py
```diff
@@ -372,6 +407,19 @@ def match_old_buildid(throttler, data):
         condition=lambda throttler, x: x == "ShutDownKill",
         result=(10, CONTINUE, REJECT),
     ),
+    # Accept 25% crash reports from Firefox ESR Windows <= 8.1
+    Rule(
+        rule_name="is_unsupported_firefox_esr_windows_version",
```
willkg, line 412: We should change this rule name to `is_firefox_esr_unsupported_windows` which reads better.

PR #946, opened 2023-10-18
Fix script to send example crash report.

smarnach:
Since the script uses `set -u`, undefined variables are an error, so the default value for `$1` doesn't work. If no command-line argument is provided, the script errors out immediately. This change makes the default value work as expected.

### DIFF
--- a/bin/send_crash_report.sh
+++ b/bin/send_crash_report.sh
@@ -5,12 +5,7 @@
 
 set -euo pipefail
 
-HOST=$1
-if [ -z "${HOST}" ]; then
-    HOST="http://localhost:8000"
-fi
-
-URL="${HOST}/submit"
+URL="${1:-http://localhost:8000}/submit"
 
 curl -v -H 'Host: crash-reports' \
      -F 'uuid=a448814e-16dd-45fb-b7dd-b0b522161010' \


smarnach:
Looks like CI is failing when trying to push to Docker Hub. It certainly shouldn't try pushing to Docker Hub in a PR build.

bin/send_crash_report.sh
```diff
@@ -5,16 +5,9 @@
...
-
-URL="${HOST}/submit"
-
 curl -v -H 'Host: crash-reports' \
      -F 'uuid=a448814e-16dd-45fb-b7dd-b0b522161010' \
      -F 'ProductName=Firefox' \
      -F 'Version=1' \
      -F upload_file_minidump=@tests/data/fakecrash.dump \
-     "$URL"
+     "${1:-http://localhost:8000}/submit"
```
willkg, line 13: This is definitely an improvement! Plus I bet we can do similar things in a bunch of other scripts.

However, I'd prefer not burying defaults. Can we set this value to a variable and then use the variable here?

```shell
URL="${1:-http://localhost:8000}/submit"

curl -v -H 'Host: crash-reports' \
     -F 'uuid=a448814e-16dd-45fb-b7dd-b0b522161010' \
     -F 'ProductName=Firefox' \
     -F 'Version=1' \
     -F upload_file_minidump=@tests/data/fakecrash.dump \
     "${URL}"
```

willkg: Also, this script isn't great. It's missing a Usage line and some other things I've been doing more recently. Plus it's not very flexible--it hard-codes uuid and the location of the dump. I wonder if I have another script for sending crash reports somewhere that I've been using instead. I haven't touched Antenna in a while. I'll have to look.

willkg (changes requested):

bin/send_crash_report.sh
smarnach: The script is referenced in the [setup quickstart](https://antenna.readthedocs.io/en/latest/dev.html#setup-quickstart), and I think it's good enough to verify that the dev environment is working fine. Everything can be improved, but I think this has value, so let's just keep it and improve it once we need it.

My first version of this change looked exactly like the code you are proposing, and then I decided this script is simple enough to inline `URL`, but I think you are right – I'll revert that.

willkg (approved):
This looks good! Thank you!

We should wait to land this until CI is fixed. [bug 1859776](https://bugzilla.mozilla.org/show_bug.cgi?id=1859776)

PR #947, opened 2023-10-18
bug 1859776: prevent Docker Hub upload script from crashing for non-main branch builds.

smarnach:
The upload script currently errors out for builds that are neither on the main branch nor for a git tag, since the CIRCLE_TAG variable will be undefined in these cases, and we use `set -u` to tell bash to error out on undefined variables. The script also errors out when DOCKERHUB_REPO is not set. This change makes the script more robust against undefined variables by checking every variable before it is used and returning with a success error code when no image upload is required.

I didn't test this locally, since I figured it will be tested automatically when CI runs.

### DIFF
--- a/bin/circleci_push.sh
+++ b/bin/circleci_push.sh
@@ -11,7 +11,7 @@
 # Environment variables used:
 #
 # DOCKER_USER/DOCKER_PASS: dockerhub credentials
-# DOCKER_REPO: name of the repo on dockerhub to use
+# DOCKERHUB_REPO: name of the repo on dockerhub to use
 # CIRCLE_BRANCH/CIRCLE_TAG: CircleCI generated environment variables
 
 set -euo pipefail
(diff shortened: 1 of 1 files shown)


willkg:
We've got development conventions for the commit summary, PR summary, and how to attach things to the relevant Bugzilla bug. We haven't talked about those, yet--that's on me.

Documentation for that in Antenna is here and I think it's up-to-date:

https://antenna.readthedocs.io/en/latest/dev.html#code-workflow

Can you rework this PR to support those?

smarnach:
I updated the title and the bug description – please take another look!

willkg (changes requested):
One last thing--can you fix the commit message summary so it includes the bug number?:

This is what it looks like now:

image

It should be more like:

```
bug 1859776: Make Docker Hub upload script more robust.
```

Everything else looks great! It might be worth going through all the bash scripts and improving them.

smarnach:
> can you fix the commit message summary so it includes the bug number?

Done.

> It might be worth going through all the bash scripts and improving them.

We should first consider whether we want to migrate to GitHub Actions, since that the direction we are moving. It would also make it easy to createn a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) for all our services (or at least the ones that are similar enough), so we wouldn't need to fix bugs like this one in multiple places.

willkg:
We never explicitly talked about "once a PR is approved, who lands it?" I think we should go with "the author lands the PR". That makes it easier for the author to handle complicated situations where multiple PRs might need to get landed and deployed in a specific sequence.

When you're ready to land this, go for it! I think I have "rebase and merge" and "squash and merge" both enabled--either is fine with me.

PR #962, opened 2024-01-17
bug-1819628: reject incoming crash reports by Android_PackageName

willkg:
This adds a map of product -> supported `Android_PackageName` values and a throttler rule to reject incoming crash reports where the configuration denotes a list of supported packagenames and either the `Android_PackageName` is missing or it's not in the supported list.

The initial set of supported packagenames is provided in bug-1819628.

### DIFF
--- a/antenna/throttler.py
+++ b/antenna/throttler.py
@@ -100,6 +100,11 @@ class Throttler:
             doc="Python dotted path to list of supported products",
             parser=parse_attribute,
         )
+        product_packagenames = Option(
+            default="antenna.throttler.MOZILLA_PRODUCT_PACKAGENAMES",
+            doc="Python dotted path to map of product -> list of supported packagenames",
+            parser=parse_attribute,
+        )
 
     def __init__(self, config):
         self.config = config.with_options(self)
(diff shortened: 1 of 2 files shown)


antenna/throttler.py
```diff
@@ -254,10 +259,32 @@ def match_unsupported_product(throttler, data):
     """Match unsupported products."""
     products = throttler.config("products")
     product_name = safe_get(data, "ProductName")
-    is_not_supported = product_name not in products
+    is_not_supported = products and product_name not in products
+
+    if is_not_supported:
+        logger.info("ProductName rejected: %r", product_name)
+        return True
+    return False
```
willkg, line 267: I cleaned up the logic here a little, but it should do the same thing it did before.

antenna/throttler.py
```diff
@@ -254,10 +259,32 @@ def match_unsupported_product(throttler, data):
...
+
+    packagenames = throttler.config("product_packagenames").get(product_name, [])
+    packagename = safe_get(data, "Android_PackageName")
 
-    if products and is_not_supported:
-        logger.info("ProductName rejected: %r" % product_name)
+    is_not_supported = packagenames and packagename not in packagenames
+
+    if is_not_supported:
+        if not packagename:
```
willkg, line 280: This isn't quite right. I need to adjust this.

antenna/throttler.py
```diff
@@ -332,6 +361,33 @@ def match_unsupported_windows(throttler, data):
 ]
 
 
+#: This accepts crash reports for all product packagenames
+ALL_PRODUCT_PACKAGENAMES = {}
+
+
+# Supported packagenames values; these have to match the Android_PackageName of the
+# incoming crash report
+MOZILLA_PRODUCT_PACKAGENAMES = {
```
smarnach, line 370: I'm wondering whether we'd prefer to have this data in a configuration file rather than in the source code, but I guess either way is fine.

smarnach (approved):
Looks good to me.

willkg:
Thank you!

PR #963, opened 2024-01-24
bug-1809500: make crashmover run synchronously

relud:


antenna/crashmover.py
```diff
@@ -137,102 +87,47 @@ def check_health(self, state):
...
-        """Spawn a crashmover if there's work to do."""
-        # Spawn a new crashmover if there's stuff in the queue and we haven't
-        # hit the limit of how many we can run
-        if self.crashmover_queue and self.crashmover_pool.free_count() > 0:
-            self.crashmover_pool.spawn(self.crashmover_process_queue)
-
-    def add_crashreport(self, raw_crash, dumps, crash_id):
-        """Add a new crash report to the crashmover queue.
+    @MYMETRICS.timer("crash_handling.time")
+    def handle_crashreport(self, raw_crash, dumps, crash_id):
```
relud, line 91: this timing was moved here from `crashmover_finish`, because synchronous handling means we don't need to manually calculate it anymore

antenna/crashmover.py
```diff
@@ -137,102 +87,47 @@ def check_health(self, state):
...
+                LOGGER.exception(
+                    f"Exception when publishing crash ({crash_id}); error {errors}/{MAX_ATTEMPTS}"
+                )
 
-        MYMETRICS.timing("crash_handling.time", value=delta)
-        MYMETRICS.incr("save_crash.count")
+        # After MAX_ATTEMPTS, we give up on this crash
+        LOGGER.error(f"{crash_id}: too many errors trying to publish; dropped")
+        MYMETRICS.incr("publish_crash_dropped.count")
+        return False
```
willkg, line 130: If we've gotten here, then the crash report data is in S3 and all we need to do is publish the crash id to the processing queue. If this fails, Socorro will notice. It's got a daily task that looks for raw crash files for which there's no corresponding processed crash file and add an entry to the missing processed crashes table. You can see that here:

https://crash-stats.mozilla.org/siteadmin/crashstats/missingprocessedcrash/

That page lets you reprocess crash ids where there isn't a processed crash report.

I thought Socorro also self-healed in that it would automatically reprocessed crash reports, but it doesn't. I forget what I was thinking in that regard. I must have been fixing it by hand for a time. There's a `MissingProcessedCrashes` API and a corresponding `fetch_missing` command, so that seems likely.

* https://crash-stats.mozilla.org/api/#MissingProcessedCrash
* https://github.com/mozilla-services/socorro/blob/main/socorro/scripts/fetch_missing.py

We should fix that to self-heal--it should put crash reports in the queue to get processed. If processing fails, it'll send Sentry events, so we'll always have some signal of crash reports failing to process. I'll write up a bug for that. It should be a straight-forward fix to `verifyprocessed.py`.

Anyhow, that's a long train of thought to get to what Antenna should do when it succeeds in saving the crash data to S3, but fails at publishing the crash id for processing. In this case, I think `handle_crashreport` should return `True` and Antenna should return an HTTP 200 with the crash id. Socorro should catch the problem and self-heal.

Also, the way this loop is written, it'll retry 20 times in rapid succession. I think it should retry 5 times waiting 2 seconds in between attempts and then give up and return `True`.

antenna/crashmover.py
```diff
@@ -137,102 +87,47 @@ def check_health(self, state):
...
-                MYMETRICS.incr("%s_crash_exception.count" % crash_report.state)
-                crash_report.errors += 1
+                MYMETRICS.incr("save_crash_exception.count")
                 LOGGER.exception(
-                    "Exception when processing queue (%s), state: %s; error %d/%d",
-                    crash_report.crash_id,
-                    crash_report.state,
-                    crash_report.errors,
-                    MAX_ATTEMPTS,
+                    f"Exception when saving crash ({crash_id}); error {errors}/{MAX_ATTEMPTS}"
```
willkg, line 108: The way this loop is implemented, it will fail 20 times in rapid succession and send a bunch of Sentry events. I think it doesn't need to try that hard (20 times is a lot) and I think it should pause between attempts. 

Previously, with the background processing crashmover, it'd get appended to the crashmover queue and there'd be some pause until it was attempted again. Antenna tried 20 times because in that situation, the crash reporter thinks the crash report has been submitted successfully and dropped the crash report data so it was important for Antenna to save it.

There's a `antenna.util.retry` you can use for this loop and the crashpublish loop.

For this loop, I think it should try 5 times with 2 seconds in between iterations. If it's not working, it should give up, return `False`, and then Antenna should return an HTTP 500 error.

willkg (changes requested):
I have some things that should get changed. After that, I'll go through and look at whether this emits the same metrics and verify some other things.

This is awesome! This removes a ton of complexity from the application and makes it a lot more straight-forward and easier to reason about. I'm excited you're working through this.

antenna/crashmover.py
```diff
@@ -2,54 +2,43 @@
...
-STATE_SAVE = "save"
-STATE_PUBLISH = "publish"
 
+def _incr_wait_generator(
+    counter, attempts=MAX_ATTEMPTS, sleep_seconds=RETRY_SLEEP_SECONDS
+):
+    def _generator_generator():
+        for _ in range(attempts - 1):
+            MYMETRICS.incr(counter)
+            yield sleep_seconds
```
relud, line 30: `MYMETRICS.incr(counter)` gets called when `antenna.util.retry` peeks a value (but only once per value, not once per peek). This is means that technically if `antenna.util.retry` peeks on the value without using it, then we would over-count, but right now it doesn't behave that way. An alternative would be specifying sleep_function and emitting this metric there.

antenna/crashmover.py
```diff
@@ -2,54 +2,44 @@
...
-        self.raw_crash = raw_crash
-        self.dumps = dumps
-        self.crash_id = crash_id
-        self.errors = errors
+def _incr_wait_generator(
+    counter, attempts=MAX_ATTEMPTS, sleep_seconds=RETRY_SLEEP_SECONDS
+):
+    def _generator_generator():
+        for _ in range(attempts - 1):
+            MYMETRICS.incr(counter)
```
relud, line 30: `MYMETRICS.incr(counter)` gets called when `antenna.util.retry` peeks a value (but only once per value, not once per peek). This is means that technically if `antenna.util.retry` peeks on the value without using it, then we would over-count, but right now it doesn't behave that way. An alternative would be specifying sleep_function and emitting this metric there.

antenna/crashmover.py
```diff
@@ -110,18 +92,24 @@ def __init__(self, config):
...
-        register_for_life(self.has_work_to_do)
+        # configure retries on save and publish
+        self.crashmover_save = retry(
+            module_logger=LOGGER,
+            wait_time_generator=_incr_wait_generator(
+                counter="save_crash_exception.count",
+                attempts=self.config("max_attempts"),
+                sleep_seconds=self.config("retry_sleep_seconds"),
+            ),
+        )(self.crashmover_save)
```
willkg, line 103: I wouldn't stomp on the original name. I'd change the new property name to something like `self.crashmover_save_with_retry` or something like that. Ditto for `crashmover_publish` below.

Another way to do this is not make `max_attempts` or `retry_sleep_seconds` configurable things and instead make them constants at the top of the module. Then you could use the retry decorator at the point the two methods are defined. I don't think these are things we're going to want to tune in server configurations.

willkg (changes requested):
If I do this:

```
make run
```

it kicks up this error:

```
antenna-web-1         | + PORT=8000
antenna-web-1         | + GUNICORN_WORKERS=1
antenna-web-1         | + GUNICORN_WORKER_CONNECTIONS=4
antenna-web-1         | + GUNICORN_WORKER_CLASS=gevent
antenna-web-1         | + GUNICORN_MAX_REQUESTS=0
antenna-web-1         | + GUNICORN_MAX_REQUESTS_JITTER=0
antenna-web-1         | + CMD_PREFIX=
antenna-web-1         | + gunicorn --workers=1 --worker-connections=4 --worker-class=gevent --max-requests=0 --max-requests-jitter=0 --config=antenna/gunicornhooks.py --log-file=- --error-logfile=- --access-logfile=- --bind 0.0.0.0:8000 antenna.wsgi:application
antenna-web-1         | 
antenna-web-1         | Error: 'antenna/gunicornhooks.py' doesn't exist
```

We need to change `bin/run_web.sh` and take out the line that looks like this:

```
  --config=antenna/gunicornhooks.py \
```

One way to test Antenna is to run it in your local dev environment and then in another shell run the systemtests.

https://github.com/mozilla-services/antenna/tree/main/systemtest

I ran Antenna from main and this branch and ran the system tests and logged the metrics and compared the two runs. Everything looks good. It was a little confusing because the order is different but that's what you get when switching from background processing to doing everything in the HTTP request handling cycle.

A few more things to fix, but otherwise this looks good!

tests/unittest/test_crashmover.py
```diff
@@ -100,41 +44,23 @@ def test_retry_storage(self, client, caplog):
...
-            f"Exception when processing queue ({crash_id}), state: save; error 14/20",
-            f"Exception when processing queue ({crash_id}), state: save; error 15/20",
-            f"Exception when processing queue ({crash_id}), state: save; error 16/20",
-            f"Exception when processing queue ({crash_id}), state: save; error 17/20",
-            f"Exception when processing queue ({crash_id}), state: save; error 18/20",
-            f"Exception when processing queue ({crash_id}), state: save; error 19/20",
-            f"Exception when processing queue ({crash_id}), state: save; error 20/20",
+            *(
+                f"CrashMover.crashmover_save: exception , retry attempt {i}"
+                for i in range(20)
```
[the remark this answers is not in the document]
willkg: Oh whoops--you fixed this while I was reviewing. I should have noticed.

antenna/crashmover.py
relud: `self.crashmover_save_with_retry` sounds good.

I'd prefer to keep this configurable so the retry tests don't have to block for 10 seconds per test.

willkg: Oh, I see that now. That makes sense. :+1:

willkg (approved):
Looks good!

When we land this and it auto-deploys to stage, we can watch the stage dashboard:

https://earthangel-b40313e5.influxcloud.net/d/mTjlP_8Zz/socorro-stage-app-metrics?orgId=1&refresh=1m

It might make sense to go over that dashboard to see if it makes sense before landing things. I haven't maintained it very well over the years.

PR #964, opened 2024-01-31
bug-1877855: Add devcontainer configuration

relud:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Eliot",
```
smarnach, line 2:
```suggestion
  "name": "Antenna",
```

Makefile
```diff
@@ -62,6 +65,15 @@ setup: my.env .docker-build  ## | Set up services.
 run: my.env .docker-build  ## | Run the webapp and services.
 	${DC} up web fakesentry
 
+.PHONY: devcontainerbuild
+devcontainerbuild: .env .docker-build .devcontainer-build  ## | Build VS Code development container.
```
smarnach, line 69: I don't think this should depend on `.devcontainer-build`, since that would be a cyclic dependency. I know we have the same for Tecken and it works, but it does look wrong, doesn't it?

Makefile
```diff
@@ -62,6 +65,15 @@ setup: my.env .docker-build  ## | Set up services.
...
+	${DC} build devcontainer
+	touch .devcontainer-build
+
+.PHONY: devcontainer
+devcontainer: .env .docker-build  ## | Run VS Code development container.
```
smarnach, line 74: Shouldn't this depend on `.devcontainer-build` instead?

Makefile
smarnach: Thinking about it, I don't see any reason why we would need the `.devcontainer-build` and `devcontainerbuild` targets at all. docker-compose will build the container when needed, and we can simply let it do its thing.

bin/run_lint.sh
```diff
@@ -17,15 +17,14 @@ PYTHON_VERSION=$(python --version)
 
 
 if [[ "${1:-}" == "--fix" ]]; then
-    echo ">>> black fix (${PYTHON_VERSION})"
-    black $FILES
+    echo ">>> ruff fix (${PYTHON_VERSION})"
+    ruff format $FILES
+    ruff check --fix $FILES
```
smarnach, line 22: Maybe this should go in a different PR? It's a bit hard to track this change down if it's mixed in with all the devcontainer stuff.

smarnach (commented):
I suggest splitting the black -> ruff change out into a separate PR.

relud:
> I suggest splitting the black -> ruff change out into a separate PR.

will do

Makefile
relud: > docker-compose will build the container when needed

when you want to pick up changes to the devcontainer's Dockerfile you have to manually rebuild, or manually rm the existing image, because docker compose will continue using the existing image even if it's out of date

relud: i got the dependencies wrong when copying from `make build`, it should be:
```suggestion
devcontainerbuild: .env  ## | Build VS Code development container.
```

smarnach: Thanks, now I understand the use case. So far I've always used the "Rebuild Container" action inside VS Code to rebuild the container if needed, but I can see that it may be convenient to do this from the command line.

smarnach (approved):
Thanks for the updates! Looks good to me.

Makefile
relud: at least for me on mac that action seems to rebuild the *container* while re-using the image, but maybe it's supposed to do both and it's just that vscode has never built the image for me, because devcontainer doesn't work at all until i manually build the image.

PR #969, opened 2024-02-01
bug-1877855: switch from black to ruff format

relud:


docs/dev.rst
```diff
@@ -332,8 +332,7 @@ To reformat the code:
 
 We're using:
 
-* `black <https://black.readthedocs.io/en/stable/>`_:  code formatting
-* `flake8 <https://flake8.pycqa.org/en/latest/>`_: linting
+* `ruff <https://docs.astral.sh/ruff/>`_: code formatting and linting
 * `bandit <https://bandit.readthedocs.io/en/latest/>`_: security linting
```
willkg, line 336: Can we drop bandit and use Ruff's bandit rules, too?

https://docs.astral.sh/ruff/rules/#flake8-bandit-s

willkg (approved):
Maybe we should update ruff to the latest which is 0.1.15 with this PR so as to pick up any ruff format fixes?

Looks good!

relud:
enabling bandit rules in ruff is throwing some errors, so i'll defer that to a separate PR

PR #970, opened 2024-02-01
bug-1809500: use gevent.sleep for crashmover retry

relud:
I just noticed that `antenna/ext/s3/connection.py` and `antenna/ext/sqs/crashpublish.py` specifically use gevent sleep, and maybe crashmover should do the same


relud:
going in the other direction and removing gevent entirely

PR #971, opened 2024-02-01
bug-1809500: drop gevent and switch gunicorn to default worker

relud:
and remove doc references to heartbeat


docs/configuration.rst
```diff
@@ -75,10 +72,7 @@ following:
 
 .. everett:option:: GUNICORN_WORKER_CLASS
    :parser: str
-   :default: "gevent"
-
-   This has to be set to ``gevent``. Antenna does some ``GeventWorker``
-   specific things and won't work with anything else.
```
willkg, line 81: I thought about adding a directive to Everett that can suss this out of bash scripts. Then we wouldn't have to manually maintain it.

willkg (approved):
Looks great!

PR #972, opened 2024-02-01
bug-1877856: update systemtest readme

relud:
as requested in https://github.com/mozilla-services/antenna/pull/965#pullrequestreview-1856417499

I intentionally left in the comment about accessing Pub/Sub because it will soon become relevant again.


systemtest/test_env.sh
```diff
@@ -49,4 +49,7 @@ esac
 echo "HOST: ${HOST}"
 echo "NGINX_TESTS: ${NGINX_TESTS}"
 
+# make sure to run systemtest even if this script is called from the git root
+cd "$(dirname "$0")/.."
```
willkg, line 53: We need pytest to run in the systemtests directory and use the `systemtests/pytest.ini` file. Otherwise it runs the unit tests which fail.

The reason CI is passing is because it's not using `test_env.sh`--it's doing the work that `test_env.sh` does.

```suggestion
cd /app/systemtest
```

It always has to run in the image, so we can hard-code the path.

willkg (changes requested):

systemtest/test_env.sh
relud:
```suggestion
cd "$(dirname "$0")"
```

relud: oops, this is copypasta, i meant to do this:
```suggestion
cd "$(dirname "$0")"
```

why does it need to run in the image? wouldn't it work on stage when using a venv?

willkg: It might, but that's not what I was supporting when I built it. That's why the instructions tell you to do `make shell` first. If we wanted to support doing a venv on the host, then we'd probably need a requirements file and walk through building the venv.

relud: fair enough. i'll hard code it.

willkg (approved):
Looks great!

PR #973, opened 2024-02-09
bug-1863063: replace bandit with ruff

relud:


bin/service-status.py
```diff
@@ -83,7 +83,7 @@ def get_config():
 
 
 def fetch(url, is_json=True):
-    resp = requests.get(url)
```
willkg, line 86: This file isn't maintained in this repo. It's maintained in https://github.com/willkg/socorro-release/ . We can make this change here, but we should make it upstream and copy it over at some point. I can take that on if you like--it's drudgery work.

testlib/mini_poster.py
```diff
@@ -234,6 +234,16 @@ def cmdline(args):
...
         ),
     )
+    parser.add_argument(
+        "--timeout",
+        dest="timeout",
+        default=30,
+        type=float,
+        help=(
+            "How many seconds to wait for the server to send data"
+            "before giving up, as a float."
```
willkg, line 244:
```suggestion
            "How many seconds to wait for the server to send data "
            "before giving up, as a float."
```

bin/service-status.py
relud: filed

PR #974, opened 2024-02-12
bug-1878423: restore google pub/sub support

relud:


antenna/ext/pubsub/crashpublish.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self, config):
+        super().__init__(config)
+
+        self.project_id = self.config("project_id")
+        self.topic_name = self.config("topic_name")
+
+        # publish messages immediately without queuing.
+        batch_settings = BatchSettings(max_messages=1)
+        self.publisher = PublisherClient(batch_settings)
```
willkg, line 57: How is this authenticating in GCP? Previously we were using a service account file.

Also, `PublisherClient` takes an optional `PublisherOptions` which lets us specify how many times to retry, timeouts, etc. The docs are kind of unhelpful and don't state what the defaults are for retry and timeout.

https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.types.PublisherOptions

It also looks like it can take an `api_endpoint` argument. We could do what we did with sqs, make that a configuration option, and then pass that in in the case of the local dev environment. It's a little more explicit than setting an environment variable is.

Alternatively, if we want to stick with the environment variable, it'd be good for this to log something along the lines of "PUBSUB_EMULATOR_HOST is set--using emulator."

systemtest/conftest.py
```diff
@@ -170,8 +210,19 @@ def list_crashids(self):
 
 
 @pytest.fixture
-def sqshelper(config):
-    """Generate and returns a PubSub helper using env config."""
+def queue_helper(config):
```
willkg, line 213: This should log which one it instantiated so it's clearer from the tests what happened.

antenna/ext/pubsub/crashpublish.py
```diff
@@ -0,0 +1,84 @@
...
+        topic_name = Option(doc="The Pub/Sub topic name to publish to.")
+
+    def __init__(self, config):
+        super().__init__(config)
+
+        self.project_id = self.config("project_id")
+        self.topic_name = self.config("topic_name")
+
+        # publish messages immediately without queuing.
+        batch_settings = BatchSettings(max_messages=1)
```
willkg, line 56: Yay! I'm so glad we don't have to redo what I did in 2019--that was nuts.

willkg (commented):
I asked some questions on slack and captured them in the bug comments. I went through the previous pubsub period (circa 2019), the project plan / requirements,  gcp pubsub docs, gcloud library API documentation, and some other things. I built the docs (and suggested some fixes). I ran antenna using SQS and also using pubsub and ran systemtests in both cases. I can't think of other things I should be doing to be thorough here, but if you have ideas, I'm game.

I had some questions about credentials and publisher options. I had a couple of minor issues. Otherwise this looks good so far!

antenna/ext/pubsub/crashpublish.py
relud: > How is this authenticating in GCP?

per https://googleapis.dev/python/google-api-core/latest/auth.html:
> - If you’re running in a Google Virtual Machine Environment (Compute Engine, App Engine, Cloud Run, Cloud Functions), authentication should “just work”.
> - If you’re developing locally, the easiest way to authenticate is using the [Google Cloud SDK](http://cloud.google.com/sdk):
>   ```
>   $ gcloud auth application-default login
>   ```
> - If you’re running your application elsewhere, you should download a [service account](https://cloud.google.com/iam/docs/creating-managing-service-accounts#creating) JSON keyfile and point to it using an environment variable:
>   ```
>   $ export GOOGLE_APPLICATION_CREDENTIALS="/path/to/keyfile.json"
>   ```

I don't think we should maintain any additional code to explicitly add to those options.

> Also, PublisherClient takes an optional PublisherOptions which lets us specify how many times to retry, timeouts, etc. The docs are kind of unhelpful and don't state what the defaults are for retry and timeout.
> 
> https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.types.PublisherOptions

the defaults appear in [google.cloud.pubsub_v1.types.PublishFlowControl](https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.types.PublishFlowControl), message limit is 1,000 and byte limit is 10,000,000 but the limit exceeded behavior is IGNORE (effectively turns off flow control). The other options for limit exceeded behavior are block (wait until there is room) and error (reject messages). Note that these are not batch settings, but whole-client settings for limiting the amount of in-flight data. This should stick to the default of IGNORE.

> It also looks like it can take an api_endpoint argument. We could do what we did with sqs, make that a configuration option, and then pass that in in the case of the local dev environment. It's a little more explicit than setting an environment variable is.

as with auth, I don't think we should maintain anything in addition to the sdk's built-in support for specifying an emulator. Specifically for pubsub, in the past setting an alternative api endpoint was insufficient for connecting to the emulator, and in fact it was impossible to specify an emulator via options without the environment variable or manually recreating a bunch of code around instantiating grpc connections.

> Alternatively, if we want to stick with the environment variable, it'd be good for this to log something along the lines of "PUBSUB_EMULATOR_HOST is set--using emulator."

will do

.circleci/config.yml
```diff
@@ -84,6 +84,14 @@ jobs:
             docker compose up --detach --wait --wait-timeout=10 ci-web
             docker compose run --rm ci shell bash -c 'cd systemtest && NGINX_TESTS=0 POST_CHECK=1 HOST=http://ci-web:8000 pytest -vv'
 
+      - run:
+          name: Run systemtest with pubsub
+          command: |
+            echo 'CRASHMOVER_CRASHPUBLISH_CLASS=antenna.ext.pubsub.crashpublish.PubSubCrashPublish' >> my.env
```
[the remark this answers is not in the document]
relud: updated to use `sed '$d' -i my.env` to remove the last line from the file

antenna/ext/pubsub/crashpublish.py
relud: > > Also, PublisherClient takes an optional PublisherOptions which lets us specify how many times to retry, timeouts, etc. The docs are kind of unhelpful and don't state what the defaults are for retry and timeout.
> > https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.types.PublisherOptions
> 
> the defaults appear in [google.cloud.pubsub_v1.types.PublishFlowControl](https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.types.PublishFlowControl), message limit is 1,000 and byte limit is 10,000,000 but the limit exceeded behavior is IGNORE (effectively turns off flow control). The other options for limit exceeded behavior are block (wait until there is room) and error (reject messages). Note that these are not batch settings, but whole-client settings for limiting the amount of in-flight data. This should stick to the default of IGNORE.

looks like i got distracted an answered the wrong question. the only information i can find is "retry transient errors indefinitely": https://github.com/googleapis/python-pubsub/blob/efbaf617bafd6d5f04c0b0f183745547a1e95f44/google/cloud/pubsub_v1/publisher/client.py#L133-L135 but I think that doesn't account for the timeout. I'm not sure.

relud: transient errors are defined in google-api-core https://googleapis.dev/python/google-api-core/latest/_modules/google/api_core/retry.html

willkg: Re: authentication, we need to document how authentication needs to be set up. This should be covered in two places:

1. in the Socorro GCP project plan because it's stuff we need to set up for the infrastructure
2. in the PubSubCrashPublish docstring in case someone else is using antenna--they need to know how to set it up

Re: retry and timeout, is it correct to say the way we've got it set up, it'll retry indefinitely and wait 120 seconds between retry attempts? Or is the timeout "times out after 120 seconds in connection or transmit" and not about retry?

relud: > Re: retry and timeout, is it correct to say the way we've got it set up, it'll retry indefinitely and wait 120 seconds between retry attempts? Or is the timeout "times out after 120 seconds in connection or transmit" and not about retry?

after rereading [the docs for the Retry class](https://googleapis.dev/python/google-api-core/latest/retry.html#google.api_core.retry.Retry) I see there are two timeouts that I was conflating.

120 seconds is the default retry timeout, and as implemented this means that no further attempts will be made after 120 seconds, but it does not interrupt an ongoing request.

the other is the rpc timeout, and is the one specified directly to [PublisherOptions](https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.types.PublisherOptions). It is implemented in [google.api_core.timeout](https://googleapis.dev/python/google-api-core/latest/_modules/google/api_core/timeout.html), and defaults to None (does not timeout).

As I understand it: the default behavior is that the client may block indefinitely during connect and transmit, and will only timeout if it's actually getting responses from the pubsub server and those responses fall into the category of "transient errors".

Given the retry handling built in to antenna, I am going to change the default behavior to disable google's retries, and set an rpc timeout of 5 seconds. I'll include it in the docs with more details, but also mentioning here: in my past experience pubsub requests take less than 150ms, and if they take even 1s things are going horribly wrong (and the mozilla data platform's ingestion edge is at risk of failure, which has happened once since their gcp migration, mitigations were put in place, but those mitigations remain unproven because it hasn't happened again)

willkg (approved):
This is great! It helps a ton that you integrated the systemtests and run them using sqs and pubsub.

The only issue I have is that looking at the CI output--there's no way to verify that the environment configuration changes actually cause antenna to be configured with pubsub in the second systemtest run. If we typo'd the `CRASHMOVER_CRASHPULBISH_CLASS` variable, we'd see the same output, but it wouldn't be using pubsub. I'm not sure how to urge pytest into spitting something else to output short of writing a pytest plugin. I'll think about that and if I think of anything, we can do a followup PR.

relud:
> The only issue I have is that looking at the CI output--there's no way to verify that the environment configuration changes actually cause antenna to be configured with pubsub in the second systemtest run. If we typo'd the `CRASHMOVER_CRASHPULBISH_CLASS` variable, we'd see the same output, but it wouldn't be using pubsub. I'm not sure how to urge pytest into spitting something else to output short of writing a pytest plugin. I'll think about that and if I think of anything, we can do a followup PR.

I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1882199 to track that work

PR #976, opened 2024-02-13
bug-1550534: google cloud storage for local dev environment

relud:
largely copied from https://github.com/mozilla-services/tecken


bin/run_setup.sh
```diff
@@ -21,6 +21,11 @@ python ./bin/s3_cli.py delete "${CRASHMOVER_CRASHSTORAGE_BUCKET_NAME}"
 python ./bin/s3_cli.py create "${CRASHMOVER_CRASHSTORAGE_BUCKET_NAME}"
 python ./bin/s3_cli.py list_buckets
 
+echo "Delete and create GCS bucket..."
+python ./bin/gcs_cli.py delete "${CRASHMOVER_CRASHSTORAGE_BUCKET_NAME}"
```
willkg, line 25: This makes me feel a little uneasy. If it were me, I'd keep a separate set of configuration for GCS than S3 and not mix them. Ditto for PubSub and SQS. 

It makes it a little easier to answer these questions:

1. What are all the things that need to be configured for GCS? This is helpful for configuring these things in stage/prod environments.
2. What are all the configuration we're no longer using? This is helpful when we're cleaning everything up after we've migrated to GCP.

Ergo, I kind of think we should be using separate configuration. I'd probably name the new ones something like `CRASHMOVER_CRASHSTORAGE_GCS_*` and `CRASHMOVER_CRASHPUBLISH_PUBSUB_*`.

bin/gcs_cli.py
```diff
@@ -0,0 +1,136 @@
...
+
+import click
+import requests
+import urllib3
+
+from google.auth.credentials import AnonymousCredentials
+from google.cloud import storage
+from google.cloud.exceptions import NotFound
+
+base_url = os.environ["STORAGE_EMULATOR_HOST"]
```
willkg, line 21: Is there any reason we can't name this more specifically as `GCS_STORAGE_EMULATOR_HOST`?

docker/config/local_dev.env
```diff
@@ -25,6 +25,9 @@ CRASHMOVER_CRASHPUBLISH_SUBSCRIPTION_NAME=local_dev_socorro_sub
 # Set Pub/Sub library to use emulator
 PUBSUB_EMULATOR_HOST=pubsub:5010
 
+# GCS settings
+STORAGE_EMULATOR_HOST=https://gcs-emulator:4443
```
willkg, line 29: Is this really supposed to be `https`? With other emulators, I usually run `http` so we don't have to deal with certificates.

docker/images/gcs-emulator/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM fsouza/fake-gcs-server:latest
```
willkg, line 1: This is tricky. I have definitely run into cases where someone builds things after I do and we end up with different versions of things and then it takes a while to figure out that whatever problem we're seeing is due to us having two different versions of things.

Can we switch this to 1.47.8?

https://hub.docker.com/layers/fsouza/fake-gcs-server/1.47.8/images/sha256-e1a52a152250dc424bc328f01f859aeb4b3b04f9e733900f2f7fcf7238972d13?context=explore

willkg (commented):
I had a few issues with variable names and using the latest tag for fake-gcs-server. Otherwise this looks good!

docker/config/local_dev.env
relud: I copied it from tecken, @biancadanforth could you answer this?

relud: to be clear, we can run this with http, per https://github.com/fsouza/fake-gcs-server/tree/main?tab=readme-ov-file#running-with-http, but tecken doesn't and i don't know why that choice was made

bin/gcs_cli.py
relud: I think it needs to be STORAGE_EMULATOR_HOST to be picked up by the google cloud storage sdk, seeing as how this value is only used for printing messages

docker/config/local_dev.env
biancadanforth: The `fake-gcs-server` is fine with HTTP but the `google-cloud-storage` library is not IIRC. See also: https://mozilla.slack.com/archives/C060L2ZA0DT/p1703197794081399 .

bin/run_setup.sh
willkg: If it were me, I'd do it because it's easier to know what to remove from `configmap.yaml`. But I'm game for whatever you're comfortable with for both GCS and PubSub config here.

docker/config/local_dev.env
relud: i got it working with http, so i'll do that here and update tecken

willkg (approved):
Looks good!

docker-compose.yml
```diff
@@ -135,10 +135,11 @@ services:
     build:
       context: docker/images/gcs-emulator
     image: local/tecken_gcs_emulator
+    command: -port 8001 -scheme http
```
willkg, line 138: :tada:

PR #977, opened 2024-02-13
bug-1881575: support storing crashes in gcs 

relud:


willkg:
This is tied to the wrong bug. This PR is one step in updating crash ingestion to use GCS instead of S3 for storage. The Socorro processor also has a crash storage for exporting crash data to Telemetry which is ultimately stored in `telemetry.socorro_crash`. That export involves storing crash data in a different S3 bucket which a DAG looks at for importing crash data and storing it in `telemetry.socorro_crash`.

I don't think we have a general bug for switching from S3 to GCS as part of the GCP migration. We need to create a new bug.

relud:
> We need to create a new bug.

filed https://bugzilla.mozilla.org/show_bug.cgi?id=1881575 and updated this PR to point there instead

willkg:
Thank you! Sorry it took me ages to notice.

antenna/ext/gcs/crashstorage.py
```diff
@@ -0,0 +1,163 @@
...
+            logger.debug(
+                "STORAGE_EMULATOR_HOST detected, connecting to emulator: %s",
+                emulator,
+            )
+            self.client = storage.Client(
+                credentials=AnonymousCredentials(),
+                project="test",
+            )
+        else:
+            self.client = storage.Client()
```
relud, line 92: unlike for pubsub in https://github.com/mozilla-services/antenna/pull/974#discussion_r1497952874 the default retry and timeout behavior for google-cloud-storage is better defined. we use two network methods in this class: `Client.get_bucket` and `Blob.upload_from_string`, both of which have a default timeout of 60 seconds.  `Blob.upload_from_string` is expected not to retry because the client assumes it is not an idempotent action. `Client.get_bucket` may retry and I think that's fine given there is a default timeout set.

tl;dr I think this client sets sane defaults for retry and timeout, and we shouldn't mess with them unless/until we see an issue.

antenna/ext/crashstorage_base.py
```diff
@@ -17,6 +19,25 @@ class Config:
     def __init__(self, config):
         self.config = config.with_options(self)
 
+    def _path_join(self, *paths):
```
willkg, line 22: Mmm... With Socorro, "crashstorage" isn't necessarily a file-based storage. In theory, it could be anything and in fact we have a crash storage for Elasticsearch and one for Telemetry which puts things in different places. I'm on the fence about whether we want to keep that thinking for Antenna or whether it makes it harder to share the GCS code between Antenna and Socorro.

I think if it were me, I wouldn't move these methods to `CrashStorageBase` but instead either create a `FileCrashStorageMixin` or duplicate the code between the S3 and GCS classes.

antenna/ext/crashstorage_base.py
```diff
@@ -17,6 +19,25 @@ class Config:
...
+
+    def _get_dump_names_path(self, crash_id):
+        return self._path_join("v1", "dump_names", crash_id)
+
+    def _get_dump_name_path(self, crash_id, dump_name):
+        # NOTE(willkg): This is something that Socorro collector did. I'm not
+        # really sure why, but in order to maintain backwards compatability, we
+        # need to keep doing it.
+        if dump_name in (None, "", "upload_file_minidump"):
+            dump_name = "dump"
```
willkg, line 37: At some point, we should get rid of this and switch it to saving "None" and empty string to "upload_file_minidump". That'll reduce a bunch of "well, if the dump name is *this*, then we do this thing for historical reasons that's unexpected and confusing". I don't know if switching to GCS is a good opportunity for that, but it might be.

antenna/ext/crashstorage_base.py
relud: i'll go ahead and duplicate the code

antenna/ext/crashstorage_base.py
relud: good enough for me, i'll remove it for gcs

antenna/ext/crashstorage_base.py
willkg: Oh poop. I thought I did the "start a review" button. I didn't intend to make these one-off comments. Sorry about that.

relud: oop, i'll stop pushing changes then >_<

systemtest/conftest.py
```diff
@@ -275,19 +315,19 @@ def dump_key(self, crash_id, name):
...
+        assert key in storage_helper.list_objects(prefix=key)
 
         # Verify the dump_names file made it
         key = self.dump_names_key(crash_id)
-        assert key in s3conn.list_objects(prefix=key)
+        assert key in storage_helper.list_objects(prefix=key)
 
         # Verify the dumps made it
         for name in dumps.keys():
             key = self.dump_key(crash_id, name)
```
relud, line 329: this is failing in CI after the change to not override name from "upload_file_minidump" to "dump" in gcs, and is fixed by the following change, which I am waiting to push so as to not interfere with willkg's review:
```diff
diff --git a/systemtest/conftest.py b/systemtest/conftest.py
index 0327f77..28beb21 100644
--- a/systemtest/conftest.py
+++ b/systemtest/conftest.py
@@ -75,6 +75,12 @@ class GcsHelper:
             "bucket": self.bucket,
         }

+    def dump_key(self, crash_id, name):
+        if name in (None, ""):
+            name = "dump"
+
+        return f"v1/{name}/{crash_id}"
+
     def list_objects(self, prefix):
         """Return list of keys in GCS bucket."""
         self.logger.info('listing "%s" for prefix "%s"', self.bucket, prefix)
@@ -120,6 +126,12 @@ class S3Helper:
         client = session.client(**client_kwargs)
         return client

+    def dump_key(self, crash_id, name):
+        if name in (None, "", "upload_file_minidump"):
+            name = "dump"
+
+        return f"v1/{name}/{crash_id}"
+
     def list_objects(self, prefix):
         """Return list of keys in S3 bucket."""
         self.logger.info('listing "%s" for prefix "%s"', self.bucket, prefix)
@@ -309,12 +321,6 @@ class CrashVerifier:
     def dump_names_key(self, crash_id):
         return f"v1/dump_names/{crash_id}"

-    def dump_key(self, crash_id, name):
-        if name in (None, "", "upload_file_minidump"):
-            name = "dump"
-
-        return f"v1/{name}/{crash_id}"
-
     def verify_stored_data(self, crash_id, raw_crash, dumps, storage_helper):
         # Verify the raw crash file made it
         key = self.raw_crash_key(crash_id)
@@ -326,7 +332,7 @@ class CrashVerifier:

         # Verify the dumps made it
         for name in dumps.keys():
-            key = self.dump_key(crash_id, name)
+            key = storage_helper.dump_key(crash_id, name)
             assert key in storage_helper.list_objects(prefix=key)

     def verify_published_data(self, crash_id, queue_helper):
```

antenna/ext/gcs/crashstorage.py
```diff
@@ -0,0 +1,180 @@
...
+            dump_name=dump_name, crash_id=crash_id
+        )
+
+    def save_raw_crash(self, crash_id, raw_crash):
+        """Save the raw crash and related dumps.
+
+        .. Note::
+
+           If you're saving the raw crash and dumps, make sure to save the raw
+           crash last.
```
willkg, line 142: When/if we get rid of the AWS Lambda job that triggers on object save events for raw crash files in S3 and don't replace it with an equivalently implemented thing in GCP, then we can drop this requirement.

docker/config/local_dev.env
```diff
@@ -25,7 +25,7 @@ CRASHMOVER_CRASHPUBLISH_SUBSCRIPTION_NAME=local_dev_socorro_sub
 PUBSUB_EMULATOR_HOST=pubsub:5010
 
 # GCS settings
-CRASHMOVER_CRASHSTORAGE_GCS_BUCKET_NAME=antennabucket
+CRASHMOVER_CRASHSTORAGE_BUCKET_NAME=antennabucket
```
willkg, line 28: This is repeated on line 39. On the one hand, it makes sense to have S3 and GCS sections. On the other hand, it gets confusing if you change this one and the next one overrides it.

I'm not sure what to do about this short of having separate name spaces, but that gets complicated with how Antenna is structured configuration-wise. :(

tests/unittest/test_gcs_crashstorage.py
```diff
@@ -0,0 +1,137 @@
...
+        mock_client = MockStorageClient.return_value
+        bucket = mock_client.get_bucket.return_value
+        blob = bucket.blob.return_value
+        upload_calls = blob.upload_from_string.mock_calls
+        blob_calls = [c for c in bucket.blob.mock_calls if c not in upload_calls]
+        bucket_calls = [
+            c
+            for c in mock_client.get_bucket.mock_calls
+            if c not in blob_calls and c not in upload_calls
+        ]
```
willkg, line 62: At some point, I changed how I test things to use the local dev environment emulators rather than mocking everything out. It's much safer in the long run and not much harder except when one is testing exceptional circumstances like permissions.

I think we should do these tests more like what we have in Socorro:

https://github.com/mozilla-services/socorro/blob/main/socorro/tests/external/boto/test_crashstorage.py

Is that doable or is it tricky because we have both S3 and GCS stuff?

antenna/ext/gcs/crashstorage.py
```diff
@@ -0,0 +1,180 @@
...
+            date=get_date_from_crash_id(crash_id),
+            crash_id=crash_id,
+        )
+
+    def _get_dump_names_path(self, crash_id):
+        return f"v1/dump_names/{crash_id}"
+
+    def _get_dump_name_path(self, crash_id, dump_name):
+        if dump_name in (None, ""):
+            dump_name = "dump"
```
willkg, line 130: This is probably not what we want. Previously `None`, `""`, and `upload_file_minidump` would all get saved as `"dump"`--so they're all treated as the same thing.

With this change, we create a new scenario where we have `"upload_file_minidump"` but also have `"dump"`. So previously if the multipart/form-data item didn't have a name, we assumed it was `"upload_file_minidump"` and treated it like that. Now if it doesn't have a name, it gets treated like `"dump"` which isn't a thing.

We should change this to treat them all like `"upload_file_minidump"`.

```suggestion
            dump_name = "upload_file_minidump"
```

willkg (changes requested):
There's one thing that should get fixed, one thing we should figure out, and I think we should redo the tests (other than the permissions one) to use the emulator rather than mocking everything out.

Other than that, this looks good!

tests/unittest/test_gcs_crashstorage.py
relud: shouldn't be tricky at all, will do

docker/config/local_dev.env
relud: i'll delete this one, and add a comment on the next one:
```diff
 CRASHMOVER_CRASHSTORAGE_SECRET_ACCESS_KEY=foo
+# Used for S3 and GCS
 CRASHMOVER_CRASHSTORAGE_BUCKET_NAME=antennabucket
```

antenna/ext/gcs/crashstorage.py
relud: sounds good, i plan to do that before cutting over prod, so i'll remove this comment and the ones later about calling save_dumps first

relud:
@willkg after the change to add emulator based tests in `tests/external` i'm randomly getting this in other tests:
```
_____________ ERROR at setup of TestCrashMover.test_retry_storage ______________
  File "/usr/local/lib/python3.11/site-packages/_pytest/runner.py", line 341, in from_call
    result: Optional[TResult] = func()
                                ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/_pytest/runner.py", line 262, in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pluggy/_hooks.py", line 493, in __call__
    return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pluggy/_manager.py", line 115, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pluggy/_callers.py", line 130, in _multicall
    teardown[0].send(outcome)
  File "/usr/local/lib/python3.11/site-packages/_pytest/unraisableexception.py", line 83, in pytest_runtest_setup
    yield from unraisable_exception_runtest_hook()
  File "/usr/local/lib/python3.11/site-packages/_pytest/unraisableexception.py", line 78, in unraisable_exception_runtest_hook
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
pytest.PytestUnraisableExceptionWarning: Exception ignored in: <socket.socket fd=-1, family=2, type=2, proto=17>

  File "/usr/local/lib/python3.11/functools.py", line 519, in decorating_function
    wrapper = _lru_cache_wrapper(user_function, maxsize, typed, _CacheInfo)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ResourceWarning: unclosed <socket.socket fd=12, family=2, type=2, proto=17, laddr=('172.19.0.7', 44008), raddr=('172.19.0.4', 8125)>
```

it passes fine if I run:
```
pytest tests/external
pytest tests/unittest
```
or if I run
```
pytest tests/unittest tests/external
```
but it still happens if I comment out all references to google-cloud-storage.

I am fully lost on how this is happening

relud:
> I am fully lost on how this is happening

okay, slightly less lost now:

if I start from a clean main and run the following commands, the error shows up:

```bash
mkdir tests/external
mv tests/unittest/test_fs_crashstorage.py tests/external/
mv tests/unittest/conftest.py tests/
# correct the number of parents to the repo root
sed '20s/.parent//' -i tests/conftest.py
# change testpaths to all of tests/
sed '24s|tests/unittest/|tests/|' -i setup.cfg
pytest -x
```

so the error is coming from having `tests/external` at all, and isn't otherwise related to this PR.

tests/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+from antenna.app import reset_verify_funs  # noqa: E402
+import antenna.libmarkus  # noqa: E402
+
+
+def pytest_runtest_setup():
+    # Make sure we set up logging and metrics to sane default values.
+    setup_logging(logging_level="DEBUG", debug=True, host_id="", processname="antenna")
+    markus.configure([{"class": "markus.backends.logging.LoggingMetrics"}])
+    # prevent antenna from reconfiguring markus
+    antenna.libmarkus._IS_MARKUS_SETUP = True
```
relud, line 30: I believe I figured out the source of this PytestUnraisableExceptionWarning:

<details><summary>traceback</summary>

```
  File "/usr/local/lib/python3.11/site-packages/_pytest/runner.py", line 341, in from_call
    result: Optional[TResult] = func()
                                ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/_pytest/runner.py", line 262, in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pluggy/_hooks.py", line 493, in __call__
    return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pluggy/_manager.py", line 115, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pluggy/_callers.py", line 130, in _multicall
    teardown[0].send(outcome)
  File "/usr/local/lib/python3.11/site-packages/_pytest/unraisableexception.py", line 83, in pytest_runtest_setup
    yield from unraisable_exception_runtest_hook()
  File "/usr/local/lib/python3.11/site-packages/_pytest/unraisableexception.py", line 78, in unraisable_exception_runtest_hook
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
pytest.PytestUnraisableExceptionWarning: Exception ignored in: <socket.socket fd=-1, family=2, type=2, proto=17>

  File "/usr/local/lib/python3.11/functools.py", line 519, in decorating_function
    wrapper = _lru_cache_wrapper(user_function, maxsize, typed, _CacheInfo)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ResourceWarning: unclosed <socket.socket fd=12, family=2, type=2, proto=17, laddr=('172.19.0.7', 44008), raddr=('172.19.0.4', 8125)>
```

</details>

the socket used by markus to write to statsd wasn't being closed properly. solution: make sure antenna doesn't call `markus.configure` after it's called here.

docker/config/local_dev.env
```diff
@@ -36,6 +33,7 @@ CRASHMOVER_CRASHSTORAGE_ENDPOINT_URL=http://localstack:4566
 CRASHMOVER_CRASHSTORAGE_REGION=us-east-1
 CRASHMOVER_CRASHSTORAGE_ACCESS_KEY=foo
 CRASHMOVER_CRASHSTORAGE_SECRET_ACCESS_KEY=foo
+# Used for S3 and GCS
```
willkg, line 36: That is a much better idea than what I was thinking. :+1:

setup.cfg
```diff
@@ -21,7 +21,7 @@ max-line-length = 88
 [tool:pytest]
 addopts = -rsxX --tb=native --showlocals
 norecursedirs = .git docs bin
-testpaths = tests/unittest/
```
willkg, line 24: I don't understand the thinking behind this. Why not put all the tests in `tests/unittest/`? Why create a `tests/external/` and then shift some things around? The `tests/README.rst` suggests each directory is a different test system, but they all run when you do `make test`, so I'm not sure I understand the breakdown.

As a side note, one of the things I wanted to do was re-root all the tests to just `tests/` rather than `tests/unittest/`. I forget why I structured it originally the way it's structured.

systemtest/conftest.py
```diff
@@ -36,7 +39,7 @@ def nginx(config):
 
 @pytest.fixture
 def postcheck(config):
-    """Return whether or not we can verify the file was saved (need access to S3)"""
+    """Return whether or not we can verify the file was saved (need access to S3/GCS)"""
```
willkg, line 42: If we generalize this, then we don't have to update it in the future.

```suggestion
    """Return whether or not we can verify the file was saved (need access to storage)"""
```

tests/conftest.py
willkg: I don't understand why this would be kicking up a `ResourceWarning` now. That's puzzling. Having said that, this setup code is not great. Instead of calling `markus.configure` directly, I'd call `antenna.libmarkus.setup_metrics` which has the guard for calling it multiple times.

tests/external/test_gcs_crashstorage_with_emulator.py
```diff
@@ -0,0 +1,119 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg, line 1: Why put this test file in `tests/external/` instead of with the other tests in `tests/unittest/`?

tests/unittest/conftest.py
```diff
@@ -7,103 +7,17 @@
...
-from falcon.testing.client import TestClient
-import markus
 from markus.testing import MetricsMock
 import pytest
 
 
-# Add repository root so we can import antenna and testlib.
+# Add repository root so we can import testlib.
 REPO_ROOT = Path(__file__).parent.parent.parent
 sys.path.insert(0, str(REPO_ROOT))
```
willkg, line 16: Doesn't this and the equivalent code in `tests/conftest.py` cause `sys.path.insert` to insert the same path twice? `antenna` and `testlib` modules are both in the same directory.

willkg (commented):
I don't understand the changes you made in the tests and I'm puzzled about the new problem with markus not closing the statsd socket. Can you explain more about what you're thinking?

tests/unittest/conftest.py
relud: this won't be a concern because i'm going to put the new tests in tests/unittest and revert all changes to this file, but:

yes it does, but it's also already in the path when tests are run via `bin/run_tests.sh` because of `export PYTHONPATH=/app/:${PYTHONPATH:-}`, and leaving it in both places means they aren't dependent on eachother to continue functioning.

setup.cfg
relud: cool, i'll file a bug to make that change after this PR is done

tests/conftest.py
relud: i don't either, other than some side effect of putting a test in tests/external, which i have now reverted and tests pass, so i'll address this in a separate PR.

relud: bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1884215

willkg (approved):
Looks good!

PR #978, opened 2024-02-22
bug-1850980: Upgrade to python 3.11

relud:


.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/devcontainers/python:3.10-bullseye@sha256:a95bd841ad4b801a5e0c4ba183aaccefd8d1e65a3dab2116aa7b385bd70135e2
+FROM mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232
```
willkg, line 1: Does the devcontainer need to match the Dockerfile base image? If so, we should probably have comments in both files so future us know to update both at the same time.

setup.cfg
```diff
@@ -25,6 +25,7 @@ testpaths = tests/unittest/
 
 filterwarnings =
     error
+    ignore:'cgi' is deprecated and slated for removal in Python 3.13:DeprecationWarning:falcon
```
willkg, line 28: What's kicking up this error? We should add a comment above it just like we do for the below one.

willkg (approved):
r+wc. Looks good!

relud:
checked against https://github.com/mozilla-services/eliot/pull/114/files and noticed i forgot to update the fakesentry image. I changed it from the python slim image to specifically the slim-bullseye image to match antenna's image more closely.

willkg:
fakesentry doesn't _have_ to match, but if you had no docker state, it's a little faster to start everything up when there are fewer base images to download. Thank you for updating it!

PR #979, opened 2024-02-27
bug-1882199: Verify that systemtests are using the intended backend

relud:
parameterize the fixture and skip when the param doesn't match the configured backend, showing the tested backend as part of the test name

### DIFF
--- a/systemtest/conftest.py
+++ b/systemtest/conftest.py
@@ -209,15 +209,20 @@ class SQSHelper:
         return crashids
 
 
-@pytest.fixture
-def queue_helper(config):
+@pytest.fixture(params=["pubsub", "sqs"])
+def queue_helper(config, request):
     """Generate and returns a PubSub or SQS helper using env config."""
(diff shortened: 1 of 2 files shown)


relud:
well, this exposes that pubsub was not being tested 🙃

systemtest/conftest.py
```diff
@@ -209,23 +209,27 @@ def list_crashids(self):
...
+    configured_backend = "sqs"
     if (
         config("crashmover_crashpublish_class")
         == "antenna.ext.pubsub.crashpublish.PubSubCrashPublish"
     ):
-        logger.debug("using pubsub")
+        configured_backend = "pubsub"
+
+    if configured_backend != request.param:
+        pytest.skip(f"test requires {request.param}")
```
willkg, line 223: Clever! I like it.

willkg (approved):
Looks great! Thank you for doing this!

PR #980, opened 2024-02-29
Update release.py and service-status.py

willkg:
Picks up these changes:

```
d19f45b Extract bug numbers from commits and list them when using make-bug (#35)
2af6518 Update github actions workflow
76121fb Add test scaffolding
f351b04 Update linting and reformatting to use ruff
850fab2 Print url to tag information GitHub (#45)
8d0d645 fix ruff S113 Probable use of requests call without timeout
f6090eb Fix service-status.py so it works without requests (#42)
```

### DIFF
--- a/bin/service-status.py
+++ b/bin/service-status.py
@@ -88,7 +88,10 @@ def fetch(url, is_json=True):
     errors if it's not valid JSON.
 
     """
-    fp = urlopen(url, timeout=5)
+    if not url.startswith(("http:", "https:")):
+        raise ValueError("URL must start with 'http:' or 'https:'")
+    # NOTE(willkg): ruff S310 can't determine whether we've validated the url or not
+    fp = urlopen(url, timeout=5)  # noqa: S310
     data = fp.read()
     if is_json:
         return json.loads(data)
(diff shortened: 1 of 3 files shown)


bin/service-status.py
```diff
@@ -83,20 +82,17 @@ def get_config():
...
-        print(url)
-        print(f"{resp.status_code}, {resp.content}")
-        raise Exception("Bad return code")
+    """Fetch data from a url
+
+    This raises URLError on HTTP request errors. It also raises JSONDecode
+    errors if it's not valid JSON.
+
+    """
+    fp = urlopen(url, timeout=5)
```
relud, line 91: this needs to suppress S310 or perhaps in pyproject.toml
```suggestion
    fp = urlopen(url, timeout=5)  # noqa: S310
```

relud (approved):
r+wc

willkg:
^^^ I just fixed the S310 issue. I also wrote up https://github.com/willkg/socorro-release/issues/50 so we don't keep hitting these.

willkg:
Thank you!

PR #984, opened 2024-03-08
bug-1884408: Make ports exposed in docker compose configurable

relud:
see also https://github.com/mozilla-services/eliot/pull/113 and https://github.com/mozilla-services/tecken/pull/2858


biancadanforth (approved):
Looks the same as what you did in Tecken: https://github.com/mozilla-services/tecken/pull/2858 .

PR #986, opened 2024-03-20
bug-1886536: remove setup.cfg & setup.py in favor of pyproject.toml

relud:


willkg (approved):
I'm pretty sure this is fine. I think we'll know for sure after the stage deploy.

PR #994, opened 2024-04-11
bug-1888201: capture user-agent and put it in metadata

biancadanforth:
Because:
* Crash reporters specify a user-agent HTTP header in the HTTP POST request when submitting crash reports.
* The Firefox crash reporter was just updated to use a Rust re-write implementation, and we'd like to know which reports are submitted by this new crash reporter.
* The new crash reporter has a unique user-agent per https://bugzilla.mozilla.org/show_bug.cgi?id=1888071#c7.

This commit:
* Captures the user-agent in crash report metadata.

### DIFF
--- a/tests/unittest/test_breakpad_resource.py
+++ b/tests/unittest/test_breakpad_resource.py
@@ -464,3 +464,22 @@ class TestBreakpadSubmitterResourceIntegration:
         # Extract the uuid from the response content and verify that it's the
         # crash id we sent
         assert result.content.decode("utf-8") == f"CrashID=bp-{crash_id}\n"
+
+    def test_add_user_agent_to_metadata(self, client):
+        expected_user_agent = "wow"
+        crash_id = "de1bb258-cbbf-4589-a673-34f800160918"
+        data, headers = multipart_encode(
+            {
+                "uuid": crash_id,
+                "ProductName": "Firefox",
+                "Version": "1.0",
+                "upload_file_minidump": ("fakecrash.dump", io.BytesIO(b"abcd1234")),
+            }
+        )
+        headers["User-Agent"] = expected_user_agent
+        client.simulate_post("/submit", headers=headers, body=data)
+
+        crashstorage = client.get_crashmover().crashstorage
+        report = crashstorage.load_crash(crash_id)
+
+        assert report.raw_crash["metadata"]["user_agent"] == expected_user_agent
--- a/tests/unittest/test_breakpad_resource.py
+++ b/tests/unittest/test_breakpad_resource.py
@@ -439,6 +439,7 @@ class TestBreakpadSubmitterResourceIntegration:
                 "payload_compressed": "0",
                 "payload_size": 632,
                 "throttle_rule": "is_nightly",
+                "user_agent": ANY,
             },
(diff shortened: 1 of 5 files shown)


tests/unittest/test_breakpad_resource.py
```diff
@@ -392,6 +394,48 @@ def test_existing_uuid(self, client):
...
+        handle_crashreport = MagicMock()
+        crashmover = FakeCrashMover()
+        crashmover.handle_crashreport = handle_crashreport
+        bsp = BreakpadSubmitterResource(config=self.empty_config, crashmover=crashmover)
+        req = request_generator(
+            method="POST", path="/submit", headers=headers, body=data
+        )
+        resp = Response
+        bsp.on_post(req, resp)
+        handle_crashreport.assert_called_once_with(raw_crash=expected_raw_crash)
```
biancadanforth, line 437: This line will always fail, because there are attributes on `raw_crash` that will change every time the test runs (like `submitted_timestamp`). How can I check only for the attribute I care about, `user_agent`?

willkg: I wouldn't use a magic mock here. The crash report data will be in `crashmover.crashmover_queue` and you can assert against the contents there.

To deal with values that change, there are a couple of ways you can do it depending on what you're trying to do. For timestamps that you care about, you could use freezegun to freeze time. We're using that in Antenna already.

https://pypi.org/project/freezegun/

For other values like uuids which are randomly generated or values you don't care about, you can compare those values against `unittest.mock.ANY`.

```python
from unittest.mock import ANY

assert "muppets" == ANY  # true
```

biancadanforth: Thanks! Upon furhter discussion, willkg determined that there were some remnants of an architectural change to `crashmover` left over in the tests and some further cleanup required. That's captured in https://bugzilla.mozilla.org/show_bug.cgi?id=1891047. This PR should be rebased off that one (#995 ) once it's merged.

biancadanforth: How I can update my tests once #995 merges: https://github.com/mozilla-services/antenna/pull/995/files#r1561495246

Example using `ANY` in `pytest`: https://github.com/mozilla-services/antenna/pull/995/files#r1561729288

biancadanforth:
Hmm looking at test failures. Seems to be other test cases than the one I added which is interesting...

tests/unittest/test_fs_crashstorage.py
```diff
@@ -83,7 +85,8 @@ def nix_tmpdir(fn):
             + b'"payload": "multipart", '
             + b'"payload_compressed": "0", '
             + b'"payload_size": 645, '
-            + b'"throttle_rule": "accept_everything"'
+            + b'"throttle_rule": "accept_everything", '
+            + b'"user_agent": "wow"'
```
biancadanforth, line 89: I couldn't figure out how to use `ANY` here from `unittest.mock`, since this is a binary string, so I hardcoded a known header value. Open to alternatives if there's a better way.

willkg (approved):
Looks good!

PR #995, opened 2024-04-11
bug-1891047: cleanup crashmover-related code

willkg:
- bug-1891047: fix CrashStorageBase method
- bug-1891047: change NoOpCrashStorage to store entire crash report
- bug-1891047: implement load_crash in crashstorage
- bug-1891047: clean up test_breakpad_resource.py


antenna/ext/crashstorage_base.py
```diff
@@ -8,6 +8,10 @@
 logger = logging.getLogger(__name__)
 
 
+class CrashIDNotFound(Exception):
+    pass
```
willkg, line 12: As I was adding new things to crashstorage, I cribbed from socorro.

antenna/ext/gcs/crashstorage.py
```diff
@@ -170,3 +184,33 @@ def save_crash(self, crash_report):
...
+
+    def load_crash(self, crash_id):
+        """Load crash data."""
+        raw_crash = {}
+        dumps = {}
+
+        raw_crash_key = self._get_raw_crash_path(crash_id)
+        try:
+            raw_crash = json.loads(self._load_file(raw_crash_key))
+        except Exception as exc:
```
willkg, line 196: `load_file` will only be used in tests. I don't know what exceptions get raised by GCS library, so I left this general.

relud: it raises `google.cloud.exceptions.NotFound`

tests/unittest/test_breakpad_resource.py
```diff
@@ -367,13 +360,93 @@ def text_extract_payload_kvpairs_and_json(self, request_generator, metricsmock):
...
+            {"TelemetryServerURL": "ou812"},
+            {
+                "metadata": {
+                    "collector_notes": ["Removed TelemetryServerURL from raw crash."]
+                }
+            },
+        ),
+    ],
+)
+def test_cleanup_crash_report(raw_crash, expected):
```
willkg, line 394: This was part of the monolithic `TestBreakpadSubmitterResource`. I broke that up into several groups. This test and the next test are better at the top level on their own.

tests/unittest/test_breakpad_resource.py
```diff
@@ -367,13 +360,93 @@ def text_extract_payload_kvpairs_and_json(self, request_generator, metricsmock):
...
+
+def test_get_throttle_result(client):
+    raw_crash = {"ProductName": "Firefox", "ReleaseChannel": "nightly"}
+
+    bsp = BreakpadSubmitterResource(config=EMPTY_CONFIG, crashmover=FakeCrashMover())
+    assert bsp.get_throttle_result(raw_crash) == (ACCEPT, "is_nightly", 100)
+
+
+class TestBreakpadSubmitterResourceIntegration:
+    def test_submit_crash_report(self, client):
```
willkg, line 408: This test shows submitting something, then loading the crash data from crash storage and asserting things against it.

tests/unittest/test_fs_crashstorage.py
```diff
@@ -102,3 +104,70 @@ def nix_tmpdir(fn):
             ]
             == b"abcd1234"
         )
+
+    @freeze_time("2011-09-06 00:00:00", tz_offset=0)
```
willkg, line 108: September 6th, 2011 was my start date.

antenna/ext/gcs/crashstorage.py
willkg: Thank you!

relud: for consistency with s3, i think that `self._load_file` should catch `google.cloud.exceptions.NotFound` and raise `KeyNotFound`, and `KeyNotFound` should be caught here. If we then add `KeyNotFound` as a class attribute here and on `S3CrashStorage` it will be accessible for things that take either s3 or gcs.

willkg:
CI is kicking up an error about how the s3 bucket doesn't exist. That relates to another problem I had with the test configuration. I'll fix those in a bit.

relud (approved):
r+ with or without my comments

tests/unittest/test_crashstorage.py
```diff
@@ -38,9 +38,11 @@ def test_flow(self, client):
         # there's only one crash in it and then verify the contents of the crash.
 
         # Verify things got saved
-        assert crashmover.crashstorage.saved_things == [
-            {"crash_id": "de1bb258-cbbf-4589-a673-34f800160918"}
+        crash_ids = [
+            crash_report.crash_id
+            for crash_report in crashmover.crashstorage.saved_things
```
biancadanforth, line 43: Should there also be a test for the noop `load_crash` method?

tests/unittest/test_fs_crashstorage.py
```diff
@@ -102,3 +104,70 @@ def nix_tmpdir(fn):
...
+            }
+        )
+
+        # Rebuild the app the test client is using with relevant configuration.
+        client.rebuild_app(
+            {
+                "BASEDIR": str(tmpdir),
+                "BREAKPAD_THROTTLER_RULES": "antenna.throttler.ACCEPT_ALL",
+                "BREAKPAD_THROTTLER_PRODUCTS": "antenna.throttler.ALL_PRODUCTS",
+                "CRASHMOVER_CRASHSTORAGE_CLASS": "antenna.ext.fs.crashstorage.FSCrashStorage",
```
biancadanforth, line 126: Nice this is what lets you change the crashstorage backend.

antenna/ext/fs/crashstorage.py
```diff
@@ -151,3 +153,39 @@ def save_crash(self, crash_report):
...
+    def load_crash(self, crash_id):
+        """Load crash data."""
+        raw_crash = {}
+        dumps = {}
+        dump_names_fn = self._get_dump_names_path(crash_id)
+
+        try:
+            with open(dump_names_fn, "rb") as fp:
+                dump_names = json.load(fp)
+        except (IOError, OSError):
```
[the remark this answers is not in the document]
willkg: Holy moly! `IOError` isn't a thing anymore. I was thinking we should keep using `OSError` because of permission issues and other possibilities, but you're right--`FileNotFoundError` is the right thing to do here and everything else should bubble up unhandled because it suggests something is super wrong.

tests/unittest/test_gcs_crashstorage.py
```diff
@@ -158,3 +159,88 @@ def get_blob(path):
...
+                "collector_notes": [],
+                "dump_checksums": {
+                    "upload_file_minidump": "e9cee71ab932fde863338d08be4de9dfe39ea049bdafb342ce659ec5450b69ae"
+                },
+                "payload": "multipart",
+                "payload_compressed": "0",
+                "payload_size": 648,
+                "throttle_rule": "accept_everything",
+            },
+            "submitted_timestamp": ANY,
```
biancadanforth, line 180: Perfect example of what you were talking about here: https://github.com/mozilla-services/antenna/pull/994/files#r1561028547

antenna/ext/s3/connection.py
```diff
@@ -86,6 +90,8 @@ class S3Connection:
 
     """
 
+    KeyNotFound = KeyNotFound
```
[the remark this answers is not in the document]
willkg: This was a convenience thing that we do in Socorro, too. We're dumping the S3 classes soon and we don't do this in the GCS equivalents.

antenna/ext/fs/crashstorage.py
```diff
@@ -151,3 +153,39 @@ def save_crash(self, crash_report):
 
         # Save raw crash
         self.save_raw_crash(crash_id, raw_crash)
+
```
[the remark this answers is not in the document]
willkg: There are cases where it has been really handy to have `FSCrashStorage`. Over time, the `S3CrashStorage` tooling has gotten better so it's almost as easy to use `S3CrashStorage` and then `s3_cli.py` to move things around. Maybe it's worth ditching. :shrug:

antenna/ext/gcs/crashstorage.py
```diff
@@ -170,3 +185,33 @@ def save_crash(self, crash_report):
...
+        try:
+            raw_crash = json.loads(self._load_file(raw_crash_key))
+        except NotFound as exc:
+            raise CrashIDNotFound(f"{crash_id} not found") from exc
+
+        dump_names_path = self._get_dump_names_path(crash_id)
+        try:
+            dump_names = json.loads(self._load_file(dump_names_path))
+        except NotFound:
+            logger.error("raw_crash exists, but dump_names doesn't exist")
```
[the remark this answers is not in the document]
willkg: This is in the `load_crash` code and this situation would be super weird--this should never happen. I put this here because the linter was annoyed with a `pass` (which is valid). I can change it to raise an error.

antenna/ext/s3/connection.py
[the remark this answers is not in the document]
willkg: I'll fix the comment. I missed that.

This is a convenience so I don't have to import things elsewhere. It's something we do in socorro sometimes. It makes layers of abstraction a little easier to deal with. Since this is in the S3 code, it's going away soon. We don't do this in the GCS equivalent.

antenna/ext/s3/crashstorage.py
```diff
@@ -141,3 +143,33 @@ def save_crash(self, crash_report):
...
+        try:
+            raw_crash = json.loads(self.connection.load_file(raw_crash_key))
+        except self.connection.KeyNotFound as exc:
+            raise CrashIDNotFound(f"{crash_id} not found") from exc
+
+        dump_names_path = self._get_dump_names_path(crash_id)
+        try:
+            dump_names = json.loads(self.connection.load_file(dump_names_path))
+        except self.connection.KeyNotFound:
+            pass
```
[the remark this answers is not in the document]
willkg: Yes. It should be consistent across the crash storage classes. It's a weird situation that shouldn't happen and I did it differently three times. I'll fix that.

tests/unittest/test_breakpad_resource.py
```diff
@@ -58,7 +42,9 @@ def test_extract_payload(self, request_generator):
             method="POST", path="/submit", headers=headers, body=data
         )
 
-        bsp = BreakpadSubmitterResource(config=self.empty_config, crashmover=crashmover)
+        bsp = BreakpadSubmitterResource(
```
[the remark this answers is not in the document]
willkg: I'm not sure, either. This probably happened a long time ago.

tests/unittest/test_crashstorage.py
willkg: We use this in other parts of the tests so it's implicitly tested. I can add an explicit test.

bin/run_web.sh
```diff
@@ -10,15 +10,13 @@ set -euxo pipefail
 
 PORT=${PORT:-"8000"}
 GUNICORN_WORKERS=${GUNICORN_WORKERS:-"1"}
-GUNICORN_WORKER_CONNECTIONS=${GUNICORN_WORKER_CONNECTIONS:-"4"}
```
willkg, line 13: `GUNICORN_WORKER_CONNECTIONS` sets the number of coroutines for the gevent worker. We switched to the sync worker when we redid the crashmover and dropped gevent.

docs/configuration.rst
```diff
@@ -54,22 +54,6 @@ following:
    `bin/run_web.sh <https://github.com/mozilla-services/antenna/blob/main/bin/run_web.sh>`_.
 
 
-.. everett:option:: GUNICORN_WORKER_CONNECTIONS
```
willkg, line 57: We don't use this anymore.

antenna/ext/gcs/crashstorage.py
```diff
@@ -170,3 +185,27 @@ def save_crash(self, crash_report):
...
+        dumps = {}
+
+        raw_crash_key = self._get_raw_crash_path(crash_id)
+        try:
+            raw_crash = json.loads(self._load_file(raw_crash_key))
+        except NotFound as exc:
+            raise CrashIDNotFound(f"{crash_id} not found") from exc
+
+        dump_names_path = self._get_dump_names_path(crash_id)
+        dump_names = json.loads(self._load_file(dump_names_path))
```
biancadanforth, line 201: I was reviewing the past comment history, and [this comment](https://github.com/mozilla-services/antenna/pull/995#discussion_r1561749846) suggests this check for `dump_names` would be in a `try..except` block and raise an exception if no `dump_names`.

It does look like things are consistent now, with no logging or exception raised if `dump_names` is empty. I know you said it would be very unusual for this to happen. Not sure if there's any benefit to raising an exception or anything we could do on the crash ingestion side if it were to happen (other than making sure we're pulling out the files correctly).

bin/run_web.sh
biancadanforth: Did a global search, and I also see it mentioned (in the stdout example) in `dev.rst` as well if you want to remove all references.

biancadanforth (approved):
LGTM; just needs to be rebased off or merged with latest `main`. Thank you!

antenna/ext/s3/crashstorage.py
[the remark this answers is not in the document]
willkg: The `dump_names` file is created by Antenna when saving the files from a crash report--it's not part of the crash report payload.

Crash storage should save the `dump_names` file, then the dump files, then the `raw_cash` file. For example, here it is in the S3 code:

https://github.com/mozilla-services/antenna/blob/f9eff555064b4c67ceebf67f603dffd296b2dd14/antenna/ext/s3/crashstorage.py#L133-L143

Because of that, I'm not sure what scenario could happen such that there exists a `raw_crash`, but no `dump_names` when running the tests.

Generally, it's bad form to have code silently fail--it's better to log something, raise an exception, emit a metric, etc.

willkg:
Thank you!

PR #1000, opened 2024-04-29
OBS-158: Add github actions CI

relud:
add github actions matching circleci flow to push to antenna image to GAR for gcp deploy flow. CircleCI will remain until aws is gone.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -1,0 +1,93 @@
+name: Build, test and push a Docker image
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - v20[0-9][0-9].[01][0-9].[0-3][0-9]  # e.g. v2023.12.04
+      - v20[0-9][0-9].[01][0-9].[0-3][0-9]-[0-9]  # e.g. v2023.12.04-2
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    permissions:
+      contents: read
+      deployments: write
+      id-token: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Get info
+        run: |
+          uname -v
+          docker info
+      - name: Create version.json
+        run: |
+          # create a version.json per
+          # https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md
+          printf '{"commit":"%s","version":"%s","source":"%s","build":"%s"}\n' \
+          "$GITHUB_SHA" \
+          "$GITHUB_REF_NAME" \
+          "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY" \
+          "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID" > version.json
+      - name: Output version.json
+        run: cat version.json
+      - name: Build Docker images
+        run: make build
+      - name: Verify requirements.txt contains correct dependencies
+        run: |
+          docker compose run --rm --no-deps ci shell ./bin/run_verify_reqs.sh
+      - name: Run lint check
+        run: |
+          make my.env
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,93 @@
...
+          # https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md
+          printf '{"commit":"%s","version":"%s","source":"%s","build":"%s"}\n' \
+          "$GITHUB_SHA" \
+          "$GITHUB_REF_NAME" \
+          "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY" \
+          "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID" > version.json
+      - name: Output version.json
+        run: cat version.json
+      - name: Build Docker images
+        run: make build
```
relud, line 39: in antenna's circleci we log in to docker before this step, but in eliot we don't.

I'm not sure which pattern to follow here.

relud (commented):
`.github/workflows/build-and-push.yml` was copied from eliot and then steps were modified to match antenna's circleci config.

ceverything that is modified from eliot is explicitly called out in comments.

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,93 @@
...
+          echo DEPLOYMENT_ENV=prod >> "$GITHUB_ENV"
+      - name: Push the Docker image to GAR
+        if: env.IMAGE_TAG != ''
+        uses: mozilla-it/deploy-actions/docker-push@v3.9.0
+        with:
+          local_image: local/antenna_deploy_base:latest
+          image_repo_path: ${{ secrets.DOCKER_IMAGE_PATH }}
+          image_tag: ${{ env.IMAGE_TAG }}
+          workload_identity_pool_project_number: ${{ secrets.WORKLOAD_IDENTITY_POOL_PROJECT_NUMBER }}
+          project_id: ${{ secrets.GCP_PROJECT_ID }}
```
relud, line 92: I configured these secrets, but i don't know if i got them right, i will need to confirm with @smarnach when he is back

relud: figured it out

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,93 @@
+name: Build, test and push a Docker image
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - v20[0-9][0-9].[01][0-9].[0-3][0-9]  # e.g. v2023.12.04
+      - v20[0-9][0-9].[01][0-9].[0-3][0-9]-[0-9]  # e.g. v2023.12.04-2
```
willkg, line 9: Currently Antenna uses a `YYYY.MM.DD[-N]` tag scheme. We haven't changed that, yet. 

* [bug 1882879: changing Antenna tag scheme](https://bugzilla.mozilla.org/show_bug.cgi?id=1882879)
* [bug 1857938: changing Socorro tag scheme](https://bugzilla.mozilla.org/show_bug.cgi?id=1857938)

We should probably do those before setting up gha based deploys.

.github/workflows/build-and-push.yml
willkg: Antenna/AWS/CircleCI/DockerHub pushes images to Docker Hub.

For Antenna/GCP/GHA/GAR, we don't need to push images to Docker Hub, so I don't think we need to log into Docker Hub. That was the reasoning for not logging into Docker Hub with Eliot.

willkg (approved):
Except for the fact that this is changing the tag schema and we should tackle that bug first, this looks good to me.

willkg:
PR 1000! :tada:

.github/workflows/build-and-push.yml
relud: cool, we'll continue with that then, and add docker login for rate limiting only if it becomes an issue

PR #1001, opened 2024-04-29
bug-1881575: only require objectUser permissions for GCS

relud:
when deploying antenna in stage, i found it was crashing when `app.verify()` was called, and tracked that to not having get bucket permissions.

We could grant get bucket permissions, but it seems unnecessary when we could only use the permissions we need.

### DIFF
--- a/antenna/ext/gcs/crashstorage.py
+++ b/antenna/ext/gcs/crashstorage.py
@@ -103,7 +103,7 @@ class GcsCrashStorage(CrashStorageBase):
         :arg bytes data: the data to save
 
         """
-        bucket = self.client.get_bucket(self.bucket)
+        bucket = self.client.bucket(self.bucket)
         blob = bucket.blob(path)
         blob.upload_from_string(data)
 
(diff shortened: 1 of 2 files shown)


antenna/ext/gcs/crashstorage.py
```diff
@@ -126,8 +126,9 @@ def verify_write_to_bucket(self):
     def check_health(self, state):
         """Check GCS connection health."""
         try:
-            # get the bucket to verify GCS is up and we can connect to it.
-            self.client.get_bucket(self.bucket)
+            # check if a blob exists to verify GCS is up and we can connect to it,
+            # without exceeding the permissions granted by roles/storage.objectUser
+            self.client.bucket(self.bucket).blob(generate_test_filepath()).exists()
```
willkg, line 131: Will the test file always exist? I'm not sure I understand this test and what will happen in this try/except block that catches `Exception`.

relud: the test file won't exist, but that's fine because the exists function returns a boolean unless it lacks permissions

willkg (approved):
Looks good as far as I can tell.

PR #1002, opened 2024-04-30
bug-1894282: Support gcp stage in systemtest

relud:
`systemtest/test_env.sh` needs an env that points to the gcp stage url

`systemtest/test_content_length.py::TestContentLength::test_content_length_1000` needs to accept http 502 from google's load balancer, because it behaves differently from amazon load balancers.


systemtest/test_content_length.py
```diff
@@ -103,16 +103,23 @@ def test_content_length_1000(self, posturl, crash_generator, nginx):
...
+            status_code = None
 
-        # Verify we get an HTTP 504 because something timed out waiting for the
+        # Verify we get an error because something timed out waiting for the
         # HTTP client (us) to send the rest of the data which is expected
         # because we sent a bad content-length
-        assert status_code == 504
+        assert status_code in (
+            # without LB, nginx drops the connection
+            None,
```
willkg, line 115: What environment do we run the system tests in that doesn't have a load balancer, but does have nginx?

relud: when i have a spare moment: local dev

willkg: I don't understand what that means. Does that mean you wrote up a bug to change the local dev environment such that it includes nginx running in local dev?

relud: good point, i have filed a bug about it https://bugzilla.mozilla.org/show_bug.cgi?id=1894955

willkg (approved):
Looks good to me.

PR #1009, opened 2024-05-02
bug-1887640: add "cloud" to `/__version__` output

willkg:
This adds a "cloud" key to `/__version__` output so we can easily determine which cloud the service is running in.

To test:

1. run collector
2. check `http://localhost:8000/__version__` and make sure "cloud" value matches the cloud configuration you're using

### DIFF
--- a/antenna/health_resource.py
+++ b/antenna/health_resource.py
@@ -5,6 +5,7 @@
 from collections import OrderedDict
 import json
 import logging
+import os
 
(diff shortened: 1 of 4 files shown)


relud (approved):

willkg:
Oops--broke the system tests. I'll fix those.

willkg:
Thank you!

PR #1019, opened 2024-05-23
Fix logging use to be more consistent

willkg:
This makes logging use more consistent across the files. Now all module-level loggers are named LOGGER rather than a mix of LOGGER and logger.

This removes the import and logger creation in files that don't use the logger at all, too.


relud (approved):

willkg:
Thank you!

PR #1021, opened 2024-05-25
bug-1884212: re-root antenna tests

willkg:
This re-roots the antenna tests so tests are under tests/ and not tests/unittest/. The latter implies a test categorization we're not using and it's confusing.


relud (approved):

willkg:
Thank you!

PR #1041, opened 2024-07-15
bug-1908018: drop support for AWS

relud:
specifically remove support for S3 and SQS, cloud-specific test markers, "cloud" in `/__version__`, and update docs to reference gcs/pubsub instead, except `docs/spec_v1.rst` which was left untouched.

For testing i ran:
```sh
make build
make test
make setup
make run
```
and in a different shell
```sh
make shell
systemtest/test_env.sh local
```

to make sure the dev environment is correctly configured, and I think the rest is covered by CI.


willkg:
I wrote up a [bug 1908646](https://bugzilla.mozilla.org/show_bug.cgi?id=1908646) to move the contents of that spec to a Google Doc. It's awkward to maintain it over time since it defines a project specification for a specific point in time.

docs/deploy.rst
```diff
@@ -64,11 +64,11 @@ Environments
 Will this run in Heroku? Probably, but you'll need to do some Heroku footwork to
 set it up.
 
-Will this run on AWS? Yes--that's what we do.
+Will this run on GCP? Yes--that's what we do.
```
willkg, line 67: I should do a pass on removing as much of the Antenna docs as possible. There's a lot of stuff in here to help other companies/people run Antenna and we don't support that.

docs/deploy.rst
```diff
@@ -122,28 +122,28 @@ Things to check:
 3. Is the configuration correct?
 
 
-AWS S3 bucket permission issues
+Google Cloud Storage bucket permission issues
 -------------------------------
```
willkg, line 126: In reST, the header line needs to be as long as the text above it.

```suggestion
Google Cloud Storage bucket permission issues
---------------------------------------------
```

docs/overview.rst
```diff
@@ -194,12 +183,12 @@ Here are some good ones:
 
 * ``breakpad_resource.crash_save.time``
 
-  Timing. This is the time it took to save the crash to S3.
+  Timing. This is the time it took to save the crash to Google Cloud Storage.
```
willkg, line 186: This section is bunk. I want to update Antenna to Markus 5.0.0 which includes a filter that allows us to specify all the metrics in a file, enforce that only those are emitted, and then auto-document them. I'll get to that eventually. Here's the equivalent in Eliot:

https://mozilla-eliot.readthedocs.io/en/latest/metrics.html

https://github.com/mozilla-services/eliot/blob/main/eliot/eliot_metrics.yaml

tests/conftest.py
```diff
@@ -132,24 +129,6 @@ def test_foo(client, tmpdir):
     return AntennaTestClient(get_app(AntennaTestClient.build_config()))
 
 
-@pytest.fixture
-def s3mock():
```
willkg, line 136: :100:

willkg (approved):
Looks good!

I wondered about sentry_sdk integrations, but this PR doesn't block on that.

antenna/app.py
```diff
@@ -20,7 +20,6 @@
 from fillmore.scrubber import Scrubber, Rule, SCRUB_RULES_DEFAULT
 import sentry_sdk
 from sentry_sdk.integrations.atexit import AtexitIntegration
-from sentry_sdk.integrations.boto3 import Boto3Integration
```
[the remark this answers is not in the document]
relud: there is gcp integration! added per https://docs.sentry.io/platforms/python/integrations/gcp-functions/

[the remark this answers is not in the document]
relud: nevermind, that is only for cloud functions and only works on python3.7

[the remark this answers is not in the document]
relud: I believe there aren't sentry integrations for the google cloud sdk

PR #1043, opened 2024-07-24
bug-1894955: add nginx container to dev and ci environments

relud:
this enables running system tests for nginx in dev and CI.

for now, this does not publish the container for deployment in GKE.


willkg (approved):
This looks good.

I wonder if we should do something about making sure we update this nginx configuration whenever we update the one in webservices-infra. It doesn't happen often, so maybe a comment in the webservices-infra nginx configuration code to remind us to update this is good enough?

PR #1044, opened 2024-07-24
Don't use custom docker compose services for CI

relud:
circleci couldn't use volume mounts, but github actions can

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -39,20 +39,18 @@ jobs:
         run: make build
       - name: Verify requirements.txt contains correct dependencies
         run: |
-          docker compose run --rm --no-deps ci shell ./bin/run_verify_reqs.sh
+          docker compose run --rm --user="$(id --user):$(id --group)" --no-deps test shell ./bin/run_verify_reqs.sh
       - name: Run lint check
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -39,20 +39,18 @@ jobs:
         run: make build
       - name: Verify requirements.txt contains correct dependencies
         run: |
-          docker compose run --rm --no-deps ci shell ./bin/run_verify_reqs.sh
+          docker compose run --rm --user="$(id --user):$(id --group)" --no-deps test shell ./bin/run_verify_reqs.sh
```
relud, line 42: set uid/gid at runtime instead of at build time via `ANTENNA_UID`/`ANTENNA_GID` because we want to publish the image tested in CI

relud: alternatively we could define the volume mounts for `web` and `test` in `docker-compose.override.yml`, so that they are still specified as part of the default docker compose path for dev environments, and then in CI we would set `COMPOSE_FILE=docker-compose.yml` ([docs](https://docs.docker.com/compose/environment-variables/envvars/#compose_file)) to ignore `docker-compose.override.yml`, and thus remove the volume mounts.

PR #1045, opened 2024-07-25
Don't use custom docker compose services for CI

relud:
by defining volume mounts in `docker-compose.override.yml` they are used by default in the dev environment, but can be disabled for CI by setting `COMPOSE_FILE`. This reduces the differences between dev and CI, which allows CI to take advantage of commands like `make setup` and `systemtest/test_env.sh local`.

This makes `docker-compose.override.yml` no longer ignored by git, which may be undesirable. If that's not acceptable, we could use the services with volume mounts in CI, because github actions support that (unlike circleci). That would require specifying `--user="$(id --user):$(id --group)"` in CI for `docker compose run` commands that touch /app to avoid permissions issues (note: I also tested this solution, it works).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,8 +17,6 @@ services:
   # building docs.
   base:
     image: local/antenna_deploy_base
-    volumes:
-      - .:/app
 
   # Test container
   test:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -53,34 +49,6 @@ services:
     volumes:
       - .:/app
```
relud, line 50: devcontainer gets to keep the volume mount because it's not used in CI

willkg (approved):
r_plus

PR #1051, opened 2024-08-10
Fix web container to mount volume in local dev environment

willkg:
This fixes the web container so that it mounts the repo directory as `/app` which allows us to see code changes without having to rebuild.

I think what's going on is that the override for `base` doesn't get pulled into the services that extend `base`.


relud (approved):

willkg:
I'm going to land this because it only affects local dev environments.

willkg:
Thank you!

PR #1100, opened 2024-11-01
OBS-144: update from socorro-release to obs-common

relud:
also removes `gcs_cli.py` and `pubsub_cli.py` which have been moved to obs-common, but were not from socorro-release


willkg (approved):
This looks good. I made sure my local dev environment work flow works (it does).

My one concern is that by moving things out of `/app/bin`, we make those things less discoverable. How would I know what the gcs command is called unless I already knew it existed?

relud:
> My one concern is that by moving things out of `/app/bin`, we make those things less discoverable. How would I know what the gcs command is called unless I already knew it existed?

We can use the commands in a script like socorro-cmd does? If you want that lmk and i'll file a follow-up PR to do that.

willkg:
`socorro-cmd` was helpful especially when I wrote it because scripts were _all over the place_ and there were like 4 or 5 different kinds of scripts--it was nuts. In django-land, there's `manage.py` which makes it easy to know what's available. I liked that.

However, a script runner was kind of weird to maintain because there were so many different kinds of scripts that executed in different ways. Passing arguments around correctly was weird. There's some weird code everywhere to support this.

Maybe instead we could create a command that just lists the existing commands. Maybe something like this:

```
> make shell
app> python bin/listcmds.py
Available commands:
gcs_cli        pubsub_cli   fetch_crash_data   fetch_crashids      reprocess
fetch_missing  db           es                 signature
...
```

All our commands are of one of two groups:

1. installed with a Python package -- we can list these automatically because they're listed in the console_scripts entrypoint package metadata

   ```
   >>> import importlib.metadata
   >>> [ep.name for ep in importlib.metadata.entry_points()["console_scripts"] if ep.value.startswith(("obs_common", "crashstats_tools", "honcho"))]
   ["honcho", "gcs_cli", "pubsub_cli", ...]
   ```

2. located in the `/app/bin` directory

It could even be something we add to obs-common and note in our docs.

> You can launch a shell in the app container and get a list of available commands:
> ```
> > make shell
> app> listcommands
> Available commands:
> 
> honcho   gcs_cli   pubsub_cli  ...
> 
> See service developer documentation for more details.
> ```

We could build that script in one place, it gets installed everywhere, and wouldn't have to do much to maintain it.

I think I'll write this up in obs-common.

PR #1101, opened 2024-11-01
bug-1908646: move spec_v1.rst to google doc

willkg:
I moved the contents of this doc to Google Docs: https://docs.google.com/document/d/1Gfxvv0uMIAHrc-trIkCEqnvVE5w33ECWm-yPuhjJggA/edit?tab=t.0

That's a more appropriate place for it. Having it in the `docs/` created two problems:

1. when doing grep in the directory, the spec for the original antenna version kept coming up
2. sometimes people would try to update the spec, but it's for the original collector rewrite project and shouldn't be updated over time


willkg:
Thank you!

PR #1110, opened 2024-11-04
Fix dependabot updates for docker/images/*

relud:
switches from directory to directories, as documented here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories

directories explicitly supports wildcards


willkg (approved):
Nice! Thank you for fixing this!

PR #1118, opened 2024-11-14
Migrate from make to just

relud:
also switch from `ANTENNA_UID` and `ANTENNA_GID` to `USE_UID` and `USE_GID`
also switch from `my.env` to `.env`


willkg (approved):
Looks good!

PR #1127, opened 2024-11-19
OBS-144: Use docker images from obs-common

relud:


docker/images/fakesentry/Dockerfile
```diff
@@ -1,24 +1 @@
...
-
-ENV PYTHONUNBUFFERED=1 \
-    PYTHONDONTWRITEBYTECODE=1
-
-RUN pip install -U 'pip>=20' && \
-    pip install --no-cache-dir 'kent==2.0.0'
-
-USER kent
-
-ENTRYPOINT ["tini", "--", "/usr/local/bin/kent-server"]
```
smarnach, line 23: I somehow [came to the conclusion that Antenna didn't use tini](https://github.com/mozilla-services/antenna/pull/1097), while it totally did. No idea how I managed to get that wrong. It's also using tini in the main Docker image. I'll file a PR to remove that.

smarnach (approved):

PR #1143, opened 2024-12-09
bug-1933191: update architecture diagram for GCP

relud:
note: this removes the stage submitter, which is not part of antenna nor is it mentioned in the overview page where this diagram is used.


willkg:
@relud Does Antenna still use the AWS ELB or is it using a GCS GLB now? I'm pretty sure I talked with someone at some point in the last 5 years about continuing to use the AWS ELB, but I don't remember what we went with for the GCP migration.

willkg (approved):
Looks good--thank you!

PR #1186, opened 2025-02-28
obs-530: fix readthedocs configuration

willkg:
This also renames the file so it's the same across all our services and updates the OS to 22.04.


toufali (approved):

willkg:
Thank you!

PR #1199, opened 2025-03-10
obs-541: remove statsd namespace in local dev environment

willkg:
Previously, we used `STATSD_NAMESPACE` and set it in the local dev environment. We don't want to do that anymore since the application now emits full metrics keys. This removes the fake `STATSD_NAMESPACE` value from the local dev environment.

This mirrors https://github.com/mozilla-services/tecken/pull/3132 .


relud (approved):

willkg:
Thank you!

PR #1201, opened 2025-03-13
OBS-546: use git short hash instead of latest for docker image tag

relud:
https://mozilla-hub.atlassian.net/browse/OBS-546

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Set Docker image tag to the shortened commit sha for updates of the main branch
         if: github.ref == 'refs/heads/main'
         run: |
-          echo IMAGE_TAG=$(git rev-parse --short "$GITHUB_SHA") >> "$GITHUB_ENV"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -54,10 +54,10 @@ jobs:
           just run --detach --wait --wait-timeout=10 web nginx
           just shell systemtest/test_env.sh local
 
-      - name: Set Docker image tag to "latest" for updates of the main branch
+      - name: Set Docker image tag to the shortened commit sha for updates of the main branch
         if: github.ref == 'refs/heads/main'
         run: |
-          echo IMAGE_TAG=latest >> "$GITHUB_ENV"
+          echo IMAGE_TAG=$(git rev-parse --short "$GITHUB_SHA") >> "$GITHUB_ENV"
```
smarnach, line 60: I think we should specify a minimum length here to make this more predictable. The default minimum length is 7. If I run this command in the Antenna and Eliot repos, I get length 7, in Tecken I get 8 and in Socorro I get length 9. How about setting a minimum of 10?

smarnach (approved):

PR #1229, opened 2025-05-05
bug-1964356: Restrict pip to a version earlier than 25.1.

smarnach:
Use an earlier version of pip becuase of https://github.com/jazzband/pip-tools/issues/2176.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -23,7 +23,7 @@ RUN apt-get update && \
 
 COPY --chown=app:app requirements.txt /app/
 
-RUN pip install -U 'pip>=20' && \
+RUN pip install -U 'pip<25.1' && \
     pip install --no-cache-dir --no-deps --only-binary :all: -r requirements.txt && \
     pip check --disable-pip-version-check
 
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -20,6 +20,7 @@ RUN apt-get update && \
 
 # Install Python dependencies
 COPY requirements.txt /app/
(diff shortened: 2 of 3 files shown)


docker/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN apt-get update && \
 
 COPY --chown=app:app requirements.txt /app/
 
-RUN pip install -U 'pip>=20' && \
+RUN pip install -U 'pip<25.1' && \
```
willkg, line 26: Given that we want to use the latest pip generally, we probably want to include some kind of reminder to revisit this situation.

Maybe a comment in `requirements.in` near pip-tools noting that when we update pip-tools, we should revisit this pin? Or a Jira issue with a date in the summary to check the pin in a month? I did that kind of thing with data migrations.

https://bugzilla.mozilla.org/show_bug.cgi?id=1795621

glind used time-bomb tests that start failing after a certain period of time if certain conditions aren't met. That's another way to get a reminder, though it creates a situation of urgency when it goes off.

willkg (approved):
I think including a reminder to revisit pip pinning is important since we really want pip up-to-date. Reminders can come in a variety of non-code forms, so this could land as is.

docker/Dockerfile
smarnach: Thanks! I added a comment both here and in requirements.in.

I'd prefer if we could pin the pip version in a way that Depedabot understands. That would make it much easier to see when a pip upgrade breaks the build. With the current model, the build can suddenly break without any changes to the repository.

PR #1337, opened 2026-05-05
bug-2034911: Reject dump names that collide with internal folder names.

smarnach:
This makes sure dump names that collide with internally used folder names get rejected. See https://bugzilla.mozilla.org/show_bug.cgi?id=2034911 for details.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -42,6 +42,14 @@ BAD_FIELDS = [
 ]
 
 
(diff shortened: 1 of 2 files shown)


biancadanforth (approved):
LGTM! Optional to update the docs with this new error reason (it already says it's a non-exhaustive list): https://socorro.readthedocs.io/en/latest/spec_crashreport.html

Unfortunately, that's in the Socorro repo where Antenna used to live, so it's a little more irksome to update.

PR #1338, opened 2026-05-05
bug-2032276: only accept client-provided uuid from socorro-submitter

biancadanforth:
This makes sure we only accept the provided `uuid` in a crash report submission if it came from the socorro-submitter.

To ensure we continue to get crash reports submitted to socorro stage, I should only merge this after I merge the corresponding socorro-submitter PR: #mozilla-services/socorro#7197.

### DIFF
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -87,6 +87,9 @@ class BreakpadSubmitterResource:
             default="upload_file_minidump",
             doc="The name of the field in the POST data for dumps.",
         )
+        stage_submitter_bearer_token = Option(
+            default="", doc=("Stage submitter bearer auth token.")
+        )
 
     def __init__(self, config, crashmover):
         self.config = config.with_options(self)
@@ -373,9 +376,19 @@ class BreakpadSubmitterResource:
         # to generate a crash id.
         throttle_result, rule_name, percentage = self.get_throttle_result(raw_crash)
 
-        # Use a uuid if they gave us one and it's valid--otherwise create a new
-        # one.
-        if "uuid" in raw_crash and validate_crash_id(raw_crash["uuid"]):
+        # Use a uuid if provided if it's from the stage submitter
+        auth_header = req.get_header("Authorization")
+        is_from_submitter = (
+            auth_header is not None
+            and auth_header.startswith("Bearer ")
+            and auth_header.split(" ", 1)[1]
+            == self.config("stage_submitter_bearer_token")
+        )
+        if (
+            "uuid" in raw_crash
+            and is_from_submitter
+            and validate_crash_id(raw_crash["uuid"])
+        ):
             crash_id = raw_crash["uuid"]
             LOGGER.info("%s has existing crash_id", crash_id)
 
--- a/antenna/breakpad_resource.py
+++ b/antenna/breakpad_resource.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import hashlib
+import hmac
 import io
 import json
 import logging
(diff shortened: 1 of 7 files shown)


antenna/breakpad_resource.py
```diff
@@ -373,9 +376,19 @@ def on_post(self, req, resp):
...
-        # Use a uuid if they gave us one and it's valid--otherwise create a new
-        # one.
-        if "uuid" in raw_crash and validate_crash_id(raw_crash["uuid"]):
+        # Use a uuid if provided if it's from the stage submitter
+        auth_header = req.get_header("Authorization")
+        is_from_submitter = (
+            auth_header is not None
+            and auth_header.startswith("Bearer ")
+            and auth_header.split(" ", 1)[1]
+            == self.config("stage_submitter_bearer_token")
```
smarnach: This implementation is insecure by default: If we don't set `stage_submitter_bearer_token`, e.g. due to a typo in some configuration file, the configured bearer token becomes the empty string, and anyone can authenticate using `"Bearer "` as the value of the authentication header. I think we want to consider all clients unauthenticated if no bearer token is configured, so the implementation is "secure-by-default" instead.

antenna/breakpad_resource.py
```diff
@@ -373,9 +376,19 @@ def on_post(self, req, resp):
...
+        )
+        if (
+            "uuid" in raw_crash
+            and is_from_submitter
```
smarnach: This will simply ignore the provided UUID if the client doesn't authenticate as the submitter. I think we actually want to reject the crash report in that case.

smarnach (commented):
Thanks for taking this over! This looks mostly good, but I have a few comments.

I suggest two changes:

* Move the auth header check inside the existing ` if "uuid" in raw_crash…` block. If the auth check fails, reject the crash report with a new rejection reason.
* Make sure the auth check always fails if no bearer token is configured.

antenna/breakpad_resource.py
biancadanforth: Ah yes you're right. I will check first for the existence of the token from config as part of this `is_from_submitter` flag.

I also am going to improve on the strict equals check to an hmac.compare_digest as that looks to be the standard way to compare secret values that prevents timing-based brute-force attacks. Even though this is probably overkill for this endpoint. Easy one line change.

antenna/breakpad_resource.py
biancadanforth: Ah yeah I guess we don't ever expect the client to submit a uuid in the normal case in the crash report, and this would mask a misconfigured stage-submitter, leading to the problem we're trying to avoid for having the same crash IDs in stage and prod for testing. It would also be misleading to accept a crash report with a uuid and no auth when we don't actually use the provided uuid and provide no indication that we didn't use the provided uuid. Thank you!

tests/test_fs_crashstorage.py
```diff
@@ -61,9 +62,9 @@ def nix_tmpdir(fn):
         # should have written--no more and no less.
         assert sorted([nix_tmpdir(fn) for fn in files]) == sorted(
             [
-                "/antenna_crashes/20160918/raw_crash/de1bb258-cbbf-4589-a673-34f800160918.json",
-                "/antenna_crashes/20160918/dump_names/de1bb258-cbbf-4589-a673-34f800160918.json",
-                "/antenna_crashes/20160918/upload_file_minidump/de1bb258-cbbf-4589-a673-34f800160918",
+                f"/antenna_crashes/20110906/raw_crash/{crash_id}.json",
+                f"/antenna_crashes/20110906/dump_names/{crash_id}.json",
+                f"/antenna_crashes/20110906/upload_file_minidump/{crash_id}",
```
biancadanforth, line 67: This is one place we use the date encoded in the server-generated `uuid` -- as a storage partition.

tests/test_fs_crashstorage.py
```diff
@@ -84,36 +83,30 @@ def nix_tmpdir(fn):
             + b'{"upload_file_minidump": "e9cee71ab932fde863338d08be4de9dfe39ea049bdafb342ce659ec5450b69ae"}, '
             + b'"payload": "multipart", '
             + b'"payload_compressed": "0", '
-            + b'"payload_size": 645, '
+            + b'"payload_size": 483, '
```
biancadanforth, line 86: This was interesting too -- basically because I removed the `uuid` from the request payload, it decreased `payload_size`.

tests/test_breakpad_resource.py
```diff
@@ -505,26 +506,71 @@ def test_existing_uuid(self, client):
...
+                "ReleaseChannel": "nightly",
+                "upload_file_minidump": ("fakecrash.dump", io.BytesIO(b"abcd1234")),
+            }
+        )
+
+        result = client.simulate_post("/submit", headers=headers, body=data)
+        assert result.status_code == 401
+
+        # Extract the uuid from the response content and verify it's a different
+        # crash id.
```
smarnach: This comment is out of date.

tests/test_breakpad_resource.py
```diff
@@ -505,26 +506,71 @@ def test_existing_uuid(self, client):
...
+            }
+        )
+
+        headers = {**headers, "Authorization": "Bearer 123"}
+
+        result = client.simulate_post("/submit", headers=headers, body=data)
+        assert result.status_code == 400
+
+        # Extract the uuid from the response content and verify that it's the
+        # crash id we sent
```
smarnach: This comment is out of date.

smarnach (approved):
Looks good to me.

There was a surprising number of tests you needed to change. :)

PR #1350, opened 2026-08-20
Bug-2060842: Changed gzip configuration

Haseeb702:
This PR:
- Changes the gzip configurations and tests it with a payload

### DIFF
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -155,8 +155,7 @@ class AntennaApp(falcon.App):
         # This is the breakpad resource that handles incoming crash reports POSTed to
         # /submit
         self.breakpad = BreakpadSubmitterResource(
-            config=config_manager.with_namespace("breakpad"),
-            crashmover=self.crashmover,
+            config=config_manager.with_namespace("breakpad"), crashmover=self.crashmover
         )
 
     def uncaught_error_handler(self, req, resp, ex, params):
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -120,6 +120,11 @@ class AntennaApp(falcon.App):
             parser=bool,
             doc="Whether or not this is a local development environment.",
         )
+        max_decompressed_body_size = Option(
+            default=str(400 * 1024 * 1024),
+            parser=int,
+            doc="The max body decompression size limit",
+        )
         statsd_host = Option(default="localhost", doc="Hostname for statsd server.")
         statsd_port = Option(default="8125", doc="Port for statsd server.", parser=int)
         secret_sentry_dsn = Option(
@@ -155,7 +160,9 @@ class AntennaApp(falcon.App):
         # This is the breakpad resource that handles incoming crash reports POSTed to
         # /submit
         self.breakpad = BreakpadSubmitterResource(
-            config=config_manager.with_namespace("breakpad"), crashmover=self.crashmover
+            config=config_manager.with_namespace("breakpad"),
+            crashmover=self.crashmover,
+            max_decompressed_body_size=self.config("max_decompressed_body_size"),
         )
 
     def uncaught_error_handler(self, req, resp, ex, params):
--- a/antenna/app.py
+++ b/antenna/app.py
@@ -120,11 +120,6 @@ class AntennaApp(falcon.App):
             parser=bool,
             doc="Whether or not this is a local development environment.",
         )
-        max_decompressed_body_size = Option(
-            default=str(400 * 1024 * 1024),
-            parser=int,
-            doc="The max body decompression size limit",
-        )
         statsd_host = Option(default="localhost", doc="Hostname for statsd server.")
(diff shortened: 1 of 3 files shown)


antenna/breakpad_resource.py
```diff
@@ -164,7 +164,15 @@ def extract_payload(self, req):
             gzip_header = 16 + zlib.MAX_WBITS
             start_time = time.perf_counter()
             try:
-                data = zlib.decompress(req.stream.read(content_length), gzip_header)
+                MAX_DECOMPRESSED_SIZE = 400 * 1024 * 1024
```
smarnach: I suggest making this a configurable setting in app.py so we can change it without a code change. This will also allow you to either reference it in the tests, or even better: change the value for the duration of the test so we don't have to generate 400 MB filled with the letter "a". 

As a global setting, it probably should have a slightly more specific name, e.g. `max_decompressed_body_size`.

smarnach (changes requested):
This looks mostly good to me. I'd prefer we make the limit configurable with environment variables, see my comment.

I double checked that the Antenna pods have enough memory to deal with 400 megabytes of data, but that shouldn't be a problem. Currently, memory utilization of Antenna is pretty low, and the pods have 4 GB of memory.

As a final test, you could send a request with a 400 MB decompressed payload to the stage instance of Antenna.

antenna/breakpad_resource.py
Haseeb702: Thanks for the feedback. I added `max_decompressed_body_size` to the `app.py` config and changed the tests to use a smaller 8 byte value.

antenna/app.py
```diff
@@ -155,7 +160,9 @@ def __init__(self, config_manager):
         # This is the breakpad resource that handles incoming crash reports POSTed to
         # /submit
         self.breakpad = BreakpadSubmitterResource(
-            config=config_manager.with_namespace("breakpad"), crashmover=self.crashmover
+            config=config_manager.with_namespace("breakpad"),
+            crashmover=self.crashmover,
+            max_decompressed_body_size=self.config("max_decompressed_body_size"),
```
smarnach: Antenna is using Everett for handling configuration, and Everett has a [patten to pass configuration to components](https://everett.readthedocs.io/en/latest/components.html#component-architectures). This is already partially implemented here – you can see the we pass the config manager with the namespace `breakpad`. I think that means we need to prefix the configuration option name with `breakpad_`, so let's rename it to `breakpad_max_body_size`. (If we'd include the `decompressed` as well, it would get a bit too long.) Inside the constructor of `BreakpadSubmitterResource`, you should be able to extract it using `self.config("max_body_size")` or similar.

It looks like the `Config` class in `BreakpadSubmitterResource` currently contains an obsolete setting `dump_name`. Feel free to remove that when you add the `max_body_size` option.

smarnach (commented):
Thanks for the update! Antenna's configuration handling is a bit non-obvious, so I added a comment with a documentation link.

antenna/app.py
Haseeb702: This makes sense. I moved the configuration option to `breakpad_resource.py` and updated all the code that depends on it. This also simplified the test cases for me as I didn't need to pass in the body size value every test case.

antenna/breakpad_resource.py
```diff
@@ -84,19 +84,21 @@ class BreakpadSubmitterResource:
...
 
     class Config:
-        dump_field = Option(
-            default="upload_file_minidump",
-            doc="The name of the field in the POST data for dumps.",
-        )
         stage_submitter_bearer_token = Option(
             default="", doc=("Stage submitter bearer auth token.")
         )
+        breakpad_max_body_size = Option(
```
smarnach: I think this should be in app.py with this name, and without the `breakpad_` prefix here, since otherwise this won't be actually configurable.

antenna/app.py
```diff
@@ -155,7 +155,8 @@ def __init__(self, config_manager):
         # This is the breakpad resource that handles incoming crash reports POSTed to
         # /submit
         self.breakpad = BreakpadSubmitterResource(
-            config=config_manager.with_namespace("breakpad"), crashmover=self.crashmover
+            config=config_manager.with_namespace("breakpad"),
+            crashmover=self.crashmover,
```
smarnach: Can you revert the changes that only change formatting, so the patch isn't as noisy as it currently is?

Haseeb702 (commented):
As we discussed in triage, I moved the body size configuration variable from `app.py` to the `breakpad_resource.py` and removed the `breakpad_` prefix from the variable name. I also tested this patch locally by submitting a gzip payload under the size limit, which was successfully accepted by Antenna. When I submitted a payload over the size limit, it stops the decompression and rejects the request.

smarnach (approved):
Thanks for your patience with this one! This is looking good to me.

